PR #9, opened 2015-09-17
setting @mix_env at compile time

smeevil:
setting @mix_env at compile time so its available in production without including Mix at runtime.
This was an issue that came forward using Exrm as deploy mechanism.

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -1,12 +1,12 @@
 defmodule Honeybadger.Mixfile do
   use Mix.Project
-
+  @mix_env Mix.env
   def project do
     [app: :honeybadger,
      version: "0.1.2",
      elixir: "~> 1.0",
-     build_embedded: Mix.env == :prod,
-     start_permanent: Mix.env == :prod,
+     build_embedded: @mix_env == :prod,
+     start_permanent: @mix_env == :prod,
      deps: deps,
      package: package,
      name: "Honeybadger",
--- a/mix.exs
+++ b/mix.exs
@@ -1,12 +1,11 @@
 defmodule Honeybadger.Mixfile do
   use Mix.Project
-  @mix_env Mix.env
   def project do
     [app: :honeybadger,
      version: "0.1.2",
      elixir: "~> 1.0",
-     build_embedded: @mix_env == :prod,
-     start_permanent: @mix_env == :prod,
+     build_embedded: Mix.env == :prod,
+     start_permanent: Mix.env == :prod,
      deps: deps,
      package: package,
      name: "Honeybadger",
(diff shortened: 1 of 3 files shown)


smeevil:
my bad, still not working :/

rbishop:
I think the issue is with `server/0` in `Notice`. That get's evaluated at run time rather than compile time. If we do:

```elixir
current_env = Mix.env
defp server do
  %{environment_name: unquote(current_env),
    hostname: hostname,
    project_root: project_root}
end
```

That might work. Though I would think above and using a module attribute would be semantically identical.

smeevil:
Double check, it does work now , and also reports production ! :D

rbishop:
Awesome! Just to verify, is that with your change or what I proposed in my comment up above?

smeevil:
That is with my change , where setting Mix.env at compile time using @mix_env Mix.env before the function head. which was basically the same as you said afterwards :D I retracted the pull request since i thought it was not working, then realised the mix.lock had a SHA lock of the repo, and just pulled to unmodified one. once i updated the lock file and tested, everything works

rbishop:
Cool! :+1: I think the only file we need to change is `lib/honeybadger/notice.ex`. The Mixfile is only evaluted at compile time and so is `Honeybadger.notice`. Want to try and revert the changes to Mixfile and `lib/honeybadger.ex`? It should still work.

smeevil:
Indeed, the only change necessary is : 

```
  @mix_env Mix.env
  defp server do
    %{environment_name: @mix_env,
      hostname: hostname,
      project_root: project_root}
  end

```

mix.exs
```diff
@@ -1,6 +1,5 @@
 defmodule Honeybadger.Mixfile do
   use Mix.Project
-
```
rbishop: Let's add this new line back and then we're good to go!

rbishop:
Awesome! Just one other comment. Thanks for cleaning up my whitespace mess. I need to check my editor config, it appears to have gone haywire :(

smeevil:
Ah, that happens automagically :)
Using Rubymine with the elixir plugin.
It does help a lot with errors and a bit of syntax highlighting.
A lot is planned yet for proper support.

I did try to switch to spacemacs and alchemist but .......man i could not get used to it :)
I've been using Jetbrains for years while working on Ruby and C# and am to used to all the nice features and integrations. 

If you are interested you can find the plugin here : https://github.com/KronicDeth/intellij-elixir

rbishop:
Merged. Thank you for pointing this out and submitting the fix!

PR #14, opened 2015-10-14
Erlang exceptions fix

### LINKED ISSUE #13: UndefinedFunctionError: undefined function: :badarg.__struct__/0 (module :badarg is not available)
We're having an error when parsing logs using the logfmt library:

https://github.com/jclem/logfmt-elixir/issues/6

When those errors happen, we also see the following in our logs, and the error is not reported (using the `use_logger` option):

```
01:02:03.208 [warn]  Unable to notify Honeybadger! UndefinedFunctionError: undefined function: :badarg.__struct__/0 (module :badarg is not available)
```

joshuap:
This fixes #13.

### DIFF
--- a/lib/honeybadger/notice.ex
+++ b/lib/honeybadger/notice.ex
@@ -4,6 +4,7 @@ defmodule Honeybadger.Notice do
   defstruct [:notifier, :server, :error, :request]
 
   def new(exception, metadata \\ %{}, backtrace) do
+    exception = Exception.normalize(:error, exception)
     exception_mod = exception.__struct__
 
     error = %{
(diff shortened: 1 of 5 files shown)


joshuap:
[Exception.normalize](https://github.com/elixir-lang/elixir/blob/master/lib/elixir/lib/exception.ex#L85) to the *rescue*. :)

lib/honeybadger/notice.ex
```diff
@@ -4,6 +4,7 @@ defmodule Honeybadger.Notice do
   defstruct [:notifier, :server, :error, :request]
 
   def new(exception, metadata \\ %{}, backtrace) do
+    exception = Exception.normalize(:error, exception)
```
rbishop: Nice find! I did not know about that function.

rbishop:
Looks good to me :+1:

joshuap:
Thanks!

Issue #10, opened 2015-10-08
Errors during test run w/ use_logger


joshuap:
I'm getting these errors when running tests with the (extremely awesome!) `use_logger` option enabled:

```
10:24:50.457 [warn]  Unable to notify Honeybadger! UndefinedFunctionError: undefined function: :badmatch.__struct__/1 (module :badmatch is not available)
10:24:50.475 [warn]  Unable to notify Honeybadger! ArgumentError: unsupported dict: {:supervisor, {#PID<0.204.0>, :ranch_listener_sup}}
10:24:50.475 [warn]  Unable to notify Honeybadger! ArgumentError: unsupported dict: {:supervisor, {:local, Logplex.Supervisor}}
```

rbishop:
Interesting. It looks like the message for a Supervisor is a tuple instead of a Dict? That's about all I can gather from that exception. I'll play around with this tonight.

PR #18, opened 2015-10-17
Only handle error messages that are lists

rbishop:
This addresses issue #10.

Some supervisor messages were making it through and the message types
were of tuple. This caused the code that looks into the message using
Dict to fail.

This change uses a guard to check that the message is a
list and matches specifically on an error type of `:error_report`. I've
also added a catch all message type after the main handle_event to
swallow all other messages.

### DIFF
--- a/lib/honeybadger/logger.ex
+++ b/lib/honeybadger/logger.ex
@@ -16,12 +16,8 @@ defmodule Honeybadger.Logger do
     {:ok, state}
   end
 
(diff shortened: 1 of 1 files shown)


rbishop:
@joshuap can you test this out against the app where you were getting the warnings?

joshuap:
I haven't been able to reproduce the error but this looks good to me. Thanks!

Issue #1, opened 2015-07-03
Phoenix session error in Honeybadger.Plug


joshuap:
I'm probably missing something obvious (first time using Phoenix), but I am getting the error "key :session not found in: %Plug.Conn{adapter: {Plug.Adapters.Cowboy.Conn, :...}" when I try to add the plug to a vanilla controller:

https://github.com/honeybadger-io/crywolf-elixir/blob/master/web/controllers/page_controller.ex#L12

Is the issue that I need to manually configure the session in Phoenix first? I'd expect it to skip the session data if it can't access it gracefully.

rbishop:
Phoenix has some manual setup you need to do to use sessions with regard to choosing the session store. I think right now it can use either `ets` tables or cookies as the session store. Here is a link to a page with more info:

http://www.phoenixframework.org/v0.6.2/docs/sessions

One other little oddity with sessions is that Plug first requires you to call `fetch_session` on the connection before you can access session data (I think is to avoid the expense of loading that data in the event that it isn't used). You're plugging the `fetch_session` function in your router so that should be taken care of for you though.

Thanks for pointing this out! I'll add some extra handling around Phoenix apps that haven't configured a session so they fail gracefully. Do you mind pasting a stacktrace here to help me cover this?

joshuap:
@rbishop sure, here's the stacktrace:

```
        (crywolf) web/controllers/page_controller.ex:3: Crywolf.PageController.handle_errors/2
        (plug) lib/plug/error_handler.ex:94: Plug.ErrorHandler.__catch__/5
        (crywolf) lib/phoenix/router.ex:297: Crywolf.Router.dispatch/2
        (crywolf) lib/phoenix/router.ex:2: Crywolf.Router.call/2
        (crywolf) lib/crywolf/endpoint.ex:1: Crywolf.Endpoint.phoenix_endpoint_pipeline/1
        (crywolf) lib/plug/debugger.ex:90: Crywolf.Endpoint."call (overridable 3)"/2
        (crywolf) lib/phoenix/endpoint/render_errors.ex:34: Crywolf.Endpoint.call/2
        (phoenix) lib/phoenix/endpoint/cowboy_handler.ex:50: Phoenix.Endpoint.CowboyHandler.upgrade/4
```

joshuap:
[...]

Issue #8, opened 2015-09-17
Don't rely on Mix


smeevil:
When running in production i see you are relying on Mix.env but it's ill advised to include Mix as a dependency in production (according to Jose Valim and Chris McCord.

Also when using Exrm to deploy, Mix is not included in the build (unless explicitly added, which you should not:D ). Even if you do include it, Mix.env will return :dev in production.

It would be nice if you could update the dependency and in stead rely on the Application.get_env to get the values your are looking for, which as far as I could find is only getting the running environment (test/dev/prod).

I'm quite new to elixir / phoenix and am not yet sure if Application.get_env is the best way to do this, but as you are using it already extensively throughout the code base, it might be the right place. The only downside would be that you need to add the config to config/test.exs config/dev.exs config/prod.exs ?

An other option could be to set a shell environment value, but that might cause issues again for people using Heroku ?

/Gerard

rbishop:
To the best of my knowledge, we only make use of `Mix.env` at compile time so Mix should be available at that point. `notify` is a macro: https://github.com/honeybadger-io/honeybadger-elixir/blob/master/lib/honeybadger.ex#L105-L128

Has this caused an issue in production or during the release process for you?

smeevil:
In production deploying with Exrm : 

Request: GET /error
** (exit) an exception was raised:
    ** (UndefinedFunctionError) undefined function: Mix.env/0 (module Mix is not available)
        Mix.env()
        (honeybadger) lib/honeybadger/notice.ex:37: Honeybadger.Notice.server/0
        (honeybadger) lib/honeybadger/notice.ex:24: Honeybadger.Notice.new/3
        (honeybadger) lib/honeybadger.ex:136: Honeybadger.do_notify/3
        (plug) lib/plug/error_handler.ex:94: Plug.ErrorHandler.__catch__/5
        (sheep) lib/sheep/endpoint.ex:1: Sheep.Endpoint.phoenix_pipeline/1
        (sheep) lib/phoenix/endpoint/render_errors.ex:34: Sheep.Endpoint.call/2
        (plug) lib/plug/adapters/cowboy/handler.ex:15: Plug.Adapters.Cowboy.Handler.upgrade/4

rbishop:
Thanks for the stacktrace! I'll dig into this tonight.

smeevil:
[...]

PR #21, opened 2015-10-26
Get server environment from application config and add poison to application list

muhmi:
#1 Get server environment name from application config or from Mix.env - works with exrm

#2 Add poison to application list so exrm can find it

### DIFF
--- a/lib/honeybadger/utils.ex
+++ b/lib/honeybadger/utils.ex
@@ -14,4 +14,17 @@ defmodule Honeybadger.Utils do
     |> tl
     |> Enum.join(".")
   end
+
+  @doc """
+  Resolve environment name from Application configuration or from Mix.env
+  """
+  def environment_name do
+    case Application.get_env(:honeybadger, :mix_env, nil) do
+      nil ->
+        Mix.env
+      name ->
+        name
+    end
+  end
+
 end
--- a/lib/honeybadger.ex
+++ b/lib/honeybadger.ex
@@ -1,4 +1,5 @@
 defmodule Honeybadger do
+  alias Honeybadger.Utils
   alias Honeybadger.Backtrace
   alias Honeybadger.Notice
   alias HTTPoison, as: HTTP
@@ -118,7 +119,7 @@ defmodule Honeybadger do
   defp macro_notify(exception, context, stacktrace) do
     exclude_envs = Application.get_env(:honeybadger, :exclude_envs, [:dev, :test])
 
-    case Honeybadger.Utils.environment_name in exclude_envs do
+    case Utils.environment_name in exclude_envs do
       false ->
         quote do
           Task.start fn ->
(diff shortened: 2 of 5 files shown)


lib/honeybadger/utils.ex
```diff
@@ -14,4 +14,17 @@ defmodule Honeybadger.Utils do
...
     |> Enum.join(".")
   end
+
+  @doc """
+  Resolve environment name from Application configuration or from Mix.env
+  """
+  def environment_name do
+    case Application.get_env(:honeybadger, :mix_env, nil) do
+      nil ->
+        Mix.env
```
joshuap: Per #8 I'm pretty sure we shouldn't be relying on `Mix` at runtime. @rbishop can you confirm when you have the chance?

rbishop: Yes we've had a problem with that in the past. That was the original cause of #8. If you build a release Mix gets compiled out.

joshuap: OK thanks, that's what I thought.

joshuap:
@muhmi thanks for submitting; we can't merge the fix for the mix environment unfortunately because we can't depend on `Mix.env` (I know the lib currently does this, but it's something I need to change). I am working on a fix for that, and will also add the environment_name setting at that time. Sorry about that!

Regarding adding `:poison` to the application list, I think @rbishop was still looking into that. If you want you could submit a separate PR for that separately (although from what I read, we're still unsure if it's necessary).

muhmi:
ok, I will use my fork until that. It works with exrm with the changes.
Mix.env wont be called if the application config declares environment name.
Its fine to reference the module as long as you dont try to call it ;) but
I suppose it would be better not to depent on it at all.

Thanks for the quick reply!

rbishop:
Thanks! I'm currently reading through this to try and get the low down on library Applications: 

http://erlang.org/pipermail/erlang-questions/2010-April/050506.html

rbishop:
According to the documentation for Application.start/2, we should add `:poison` to the `included_applications` since it is an Application that does not have processes, or rather do not need to be started.

http://elixir-lang.org/docs/v1.0/elixir/Application.html#start/2

muhmi:
Sorry to comment on a closed request but I came across this issue with exrm:
https://github.com/bitwalker/exrm/issues/147

After I tried to move things in to `:included_applications`.

Found this bit in the comments:

"You should use applications unless you explicitly do not want the application in question to be started, but there is almost never a reason for this unless you plan to manage the start manually. Even "library" applications (I make a distinction here, but really there isn't one) can be started, though the operation is effectively a no-op. You do not want to put things in included_applications just because they do not have an application which needs to be started"

This from https://github.com/bitwalker who is the guy behind exrm. 

So I guess putting `:poison` to `:applications` is the way to go.

joshuap:
@muhmi thanks! Is that a change we still need to make in this lib? If so, could you create a new issue for it?

jonatandahora:
I'm also having problems with Mix.env always reporting as :prod as described in #8 and @muhmi fork works for me.
There isn't any way this could be merged or that you guys could add a fix that doesn't rely on Mix?

PR #22, opened 2015-10-26
Make environment name configurable.

joshuap:
This also fixes the issue of depending on Mix.env at runtime. It may *not* fix the issue of the *correct* `MIX_ENV` being picked up. I just did a test with crywolf/phoenix and it *did* default my Honeybadger environment_name to `:prod` when I ran the phoenix server with `MIX_ENV=prod`, so maybe it's not an issue if we do it this way. I think the concern was that once the honeybadger package is compiled then the default `Mix.env` is hard-coded.

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -18,6 +18,7 @@ defmodule Honeybadger.Mixfile do
 
   def application do
     [applications: [:httpoison, :logger],
+     env: [environment_name: Mix.env],
      mod: {Honeybadger, []}]
   end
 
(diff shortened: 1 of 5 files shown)


mix.exs
```diff
@@ -18,6 +18,7 @@ defmodule Honeybadger.Mixfile do
 
   def application do
     [applications: [:httpoison, :logger],
+     env: [environment_name: Mix.env],
```
rbishop: This looks safe to me. I'm at the point where the only place I'm willing to guarantee Mix exists is in the Mixfile itself :-p

rbishop:
:+1:

joshuap:
@muhmi this is my solution to #21 -- it adds the config option and sets the default value to `Mix.env` in the Mixfile. You can override it in your local config, then. :)

joshuap:
@rbishop thanks!

PR #23, opened 2015-10-27
Send the entire process dictionary as context, let Notice pull out the context

### LINKED ISSUE #19: Honeybadger.notify/2 doesn't report context.
@knewter noticed this while working on his ElixirSips screencast: `Honeybadger.notify(exception, context)` didn't report the local context with the error.

rbishop:
This fixes #19. The context wasn't getting sent because we were taking the `honeybadger_context` key out with our call to `Dict.take/2`. Instead we can just call notify with the entire process dictionary and Honeybadger.Notice does the right thing and only sends the value at `:honeybadger_context`.

### DIFF
--- a/lib/honeybadger/notice.ex
+++ b/lib/honeybadger/notice.ex
@@ -6,18 +6,13 @@ defmodule Honeybadger.Notice do
   def new(exception, metadata \\ %{}, backtrace) do
     exception = Exception.normalize(:error, exception)
     exception_mod = exception.__struct__
-
     error = %{
       class: Utils.strip_elixir_prefix(exception_mod),
       message: exception_mod.message(exception),
       tags: Dict.get(metadata, :tags, []),
       backtrace: backtrace
     }
-
-    context = Dict.get(metadata, :honeybadger_context, %{}) |> Enum.into(Map.new)
-    request = metadata
-              |> Dict.get(:plug_env, %{})
-              |> Dict.merge(%{context: context})
+    request = Dict.take(metadata, [:plug_env, :context])
 
     %__MODULE__{error: error,
                 request: request,
--- a/lib/honeybadger.ex
+++ b/lib/honeybadger.ex
@@ -153,6 +153,7 @@ defmodule Honeybadger do
 
   def context(dict) do
     Process.put(@context, Dict.merge(context, dict))
+    context
   end
 
   defp default_config do
--- a/lib/honeybadger.ex
+++ b/lib/honeybadger.ex
@@ -106,22 +106,22 @@ defmodule Honeybadger do
     macro_notify(exception, {:%{}, [], []}, [])
   end
 
-  defmacro notify(exception, context) do
-    macro_notify(exception, context, [])
+  defmacro notify(exception, metadata) do
+    macro_notify(exception, metadata, [])
   end
 
-  defmacro notify(exception, context, stacktrace) do
-    macro_notify(exception, context, stacktrace)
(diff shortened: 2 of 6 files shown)


joshuap:
@rbishop the issue on #19 (afaik) is that the local context (second argument) when calling `Honeybadger.notify(exception, context)` isn't added to the reported context. Does this address that? /cc @knewter

rbishop:
What do you mean by local context? I always thought of context as the things you add using `Honeybadger.context`.

joshuap:
It can also be added to just one exception report when calling `Honeybadger.notify` manually, can't it? It says so in the [README example](https://github.com/honeybadger-io/honeybadger-elixir#honeybadgernotify-send-an-exception-to-honeybadger).

test/honeybadger_test.exs
```diff
@@ -17,21 +18,50 @@ defmodule HoneybadgerTest do
...
+        context: %{
+          foo: "Bar"
+        }
+      },
+      server: %{
+        environment_name: :test, 
+        hostname: "rb-2",
+        project_root: "/Users/rb/code/honeybadger"
+      }
+    }
```
rbishop: I really wish there was an easier way to test the entire payload without needing all the specifics. This test will break if line numbers break. I need to think of a better way to test this.

rbishop:
@joshuap I don't know why I had such a brainfart on that. I've pushed some changes that fix that as well (and along the way I also fixed an issue with the context in the Logger :p)

rbishop:
I'm thinking this isn't passing CI because of the line numbers in that test. If the Erlang and Elixir versions are different the line numbers will be different. Back to the drawing board sooner than later on testing an entire notice.

lib/honeybadger/notice.ex
```diff
@@ -6,18 +6,13 @@ defmodule Honeybadger.Notice do
...
       message: exception_mod.message(exception),
       tags: Dict.get(metadata, :tags, []),
       backtrace: backtrace
     }
-
-    context = Dict.get(metadata, :honeybadger_context, %{}) |> Enum.into(Map.new)
-    request = metadata
-              |> Dict.get(:plug_env, %{})
-              |> Dict.merge(%{context: context})
+    request = Dict.take(metadata, [:plug_env, :context])
```
joshuap: Is this be adding a `:plug_env` key to the request?

test/honeybadger_test.exs
joshuap: Yeah, that would be nice. Makes me appreciate `kind_of(Hash)` in RSpec. :)

lib/honeybadger/notice.ex
joshuap: The reason I ask is that we don't currently support that key in our payload. We have a `:cgi_data` key which we use for CGI environment variables.

joshuap:
@rbishop makes sense. It looks good overall; I left a few other comments. Hopefully we can find a way around the test issue. If you want to skip the roundtrip notice test in the meantime that's alright with me, although it would be nice to have.

lib/honeybadger/notice.ex
rbishop: I call the key `plug_env` here because nested under that key is `url`, `session`, `component`, `params` and `cgi_data`.

rbishop:
I reverted those test payload changes so that the tests pass. This should be good to merge and fix #19

joshuap:
:+1: thanks @rbishop!

PR #33, opened 2015-12-27
Require environment_name setting for each environment

### LINKED ISSUE #31: HoneyBadger is always compiled with :prod environment
Even if i set environment_name as :dev i get notified in compile-time exceptions as prod like @joshuap described in #8.
Once the app finishes compiling, it falls back to :dev.

rbishop:
Mix automatically sets the `Mix.env` to `prod` for every library that an application depends on. Since honeybadger-elixir is always a dependency that means `environment_name` is always getting set to `prod`. Because of this, we cannot infer the current environment at compile time (or runtime because relx (release tool used by exrm) removes Mix) we need our users to explicitly set the `environment_name` per environment.

This change requires that `environment_name` is set and checks for it at Application start time and raises an exception if `environment_name` is not set. I've also added documentation to help users through this.

This fixes #31


rbishop:
@joshuap I added a check for `MIX_ENV` being present and using that if it is. Let me know what you think!

lib/honeybadger.ex
```diff
@@ -161,7 +170,26 @@ defmodule Honeybadger do
       hostname: :inet.gethostname |> elem(1) |> List.to_string,
       origin: "https://api.honeybadger.io",
       project_root: System.cwd,
-      use_logger: true,
-      environment_name: nil]
+      use_logger: true]
+  end
+
+  defp get_and_set_environment_name!(config) do
+    case System.get_env("MIX_ENV") do
```
joshuap: I'm thinking that maybe MIX_ENV should be the default, but if the honeybadger config option is set explicitly, that should be preferred. It looks like this will always prefer the environment variable. @rbishop thoughts?

rbishop: I went back and forth when I was working on this. Ultimately I ended up going with the environment variable having preference because that is more consistent with other tools that provide overridable configuration via env vars. I can see it either way though. Do the other Honeybadger clients have existing precedence with this?

joshuap: The way I do it for Ruby/Rails is:

1. RACK_ENV or RAILS_ENV environment variable (default)
2. Disk configuration (honeybadger.yml) (overrides default)
3. HONEYBADGER_ENV environment variable (always overrides disk config)

So I *do* prefer the environment variable if it belongs to the gem, but I give the least priority to 3rd party config. Does that make sense?

joshuap:
:+1: :shipit:

Issue #32, opened 2015-12-16
Do we need a top-level supervisor?


joshuap:
I tried updating honeybadger in a mix project and then releasing with exrm, and ran into an error similar to this: https://github.com/bitwalker/exrm/issues/183

Could it be that we have the same problem?

parubets:
same problem for me...

joshuap:
@parubets thanks, we'll look into this. @rbishop didn't we fix the top-level supervisor issue? Any ideas what could be going on here?

joshuap:
#37 should fix this issue. @parubets let me know if you still have the issue after updating to 0.5.0.

PR #37, opened 2016-04-18
Send web app request stats

rbishop:
This PR adds integration with the Honeybadger API for metrics. Request metrics are gathered from Phoenix and Plug apps and then sent to Honeybadger for analysis.

This PR also adds a Supervision tree which should resolve issue #32.

Another notable change is the creation of `Honeybadger.Client`. This module and its accompanying struct extract the logic for sending notices and metrics to the Honeybadger API.

### DIFF
--- a/lib/honeybadger/plug.ex
+++ b/lib/honeybadger/plug.ex
@@ -7,6 +7,22 @@ defmodule Honeybadger.Plug do
       use Plug.ErrorHandler
       require Honeybadger
 
+      def call(conn, opts) do
+        try do
+          before_response = :erlang.monotonic_time
+
+          conn = super(conn, opts)
+
+          after_response = :erlang.monotonic_time
+          response_time = :erlang.convert_time_unit(after_response - before_response, :nano_seconds, :milli_seconds)
+          Honeybadger.Metrics.Server.timing(response_time)
+          conn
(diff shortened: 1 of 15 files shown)


joshuap:
This looks really clean. Can't wait to try it. :)

lib/honeybadger/plug.ex
```diff
@@ -7,6 +7,22 @@ defmodule Honeybadger.Plug do
       use Plug.ErrorHandler
       require Honeybadger
 
+      def call(conn, opts) do
+        try do
+          before_response = :erlang.monotonic_time
+
+          conn = super(conn, opts)
+
+          after_response = :erlang.monotonic_time
```
rbishop: I just remembered that monotonic_time is only supported by Erlang/OTP version 18. The Elixir community is pretty good about using the latest versions and Elixir 1.2 and all future versions have a hard dependency on OTP 18 anyway.

test/environment_test.exs
```diff
@@ -0,0 +1,15 @@
+defmodule Honeybadger.EnvironmentTest do
```
rbishop: Whoops. These are for my branch for extracting the environment setup to a module. I'll remove from here :)

Issue #39, opened 2016-04-27
logger.ex:24: warning: variable context is unused


somlor:
Saw these warnings when compiling:
```
lib/honeybadger/logger.ex:24: warning: variable context is unused
lib/honeybadger/logger.ex:27: warning: variable exception is unused
lib/honeybadger/logger.ex:29: warning: variable exception is unused
lib/honeybadger/logger.ex:29: warning: variable stacktrace is unused
```

Forked repo and intended to submit a pull but now I'm scratching my head. Why does Elixir think that these variables are unused? 👀 
```elixir
case error_info do
  {_kind, {exception, stacktrace}, _stack} when is_list(stacktrace) ->
    Honeybadger.notify(exception, context, stacktrace)
  {_kind, exception, stacktrace} ->
    Honeybadger.notify(exception, context, stacktrace)
end
```
https://github.com/honeybadger-io/honeybadger-elixir/blob/master/lib/honeybadger/logger.ex#L24

rbishop:
This seems odd to me. Do you get this when compiling in an environment that is not in your `exclude_envs` setting for Honeybadger? We use a macro to disable reporting in excluded environments like `test` and `dev`. I wonder if that could be the cause?

somlor:
Interesting. I don't think so. I get it even when I clone this repo and run `mix test`. Is this just on my end? Let me know if I can provide anymore info. I was running Erlang 18.2.1 with Elixir 1.2.3 but I've just upgraded to 18.3 + 1.2.4 and still seeing warnings.

rbishop:
I can replicate this, too. :+1: I'll look into it.

joshuap:
@rbishop thanks!

Issue #45, opened 2016-06-24
Warnings when compiling under Elixir 1.3


derekprior:
```
==> honeybadger
Compiling 11 files (.ex)
warning: variable session is unused
  lib/honeybadger/plug.ex:50

warning: variable e is unused
  lib/honeybadger/plug.ex:52

warning: unused import Supervisor.Spec
  lib/honeybadger.ex:106

warning: variable context is unused
  lib/honeybadger/logger.ex:24

warning: variable exception is unused
  lib/honeybadger/logger.ex:27

warning: variable exception is unused
  lib/honeybadger/logger.ex:29

warning: variable stacktrace is unused
  lib/honeybadger/logger.ex:29

Generated honeybadger app
```

nurugger07:
@derekprior thanks! I'm going to look at the upgrade to Elixir v1.3 this weekend. Is everything else working well then?

derekprior:
I haven't deployed to production or staging yet, but locally it's checking out. I even have a branch where I started addressing those warnings, but the one for unused variable `context` is perplexing because it does look like it's used. Granted, I spent only 5 minutes looking at it so far.

nurugger07:
This should be resolved after the merge. I'll get @joshuap to push up a hex update :) Thanks to @mitchellhenke for the PR!

mitchellhenke:
@nurugger07 do you think a new release will be made soon?

nurugger07:
@mitchellhenke I'll be working on getting it out this week. I wanted to get a little more in :)

mitchellhenke:
@nurugger07 sounds perfect, thank you!

nurugger07:
v0.6 was released and the warnings have been cleaned up. Special thanks to @mitchellhenke :)

mitchellhenke:
@nurugger07 you bet!

PR #46, opened 2016-06-29
Fix elixir compliation warnings

mitchellhenke:
Should fix https://github.com/honeybadger-io/honeybadger-elixir/issues/39 and https://github.com/honeybadger-io/honeybadger-elixir/issues/45

I'm not sure if this is the ideal way to handle it.  I did my best to take some inspiration from how `Elixir.Logger` handles it ([link](https://github.com/elixir-lang/elixir/blob/master/lib/logger/lib/logger.ex#L626-L649)), as it's very similar to this.

Also cleaned up some other miscellaneous warnings.

### DIFF
--- a/lib/honeybadger.ex
+++ b/lib/honeybadger.ex
@@ -138,7 +138,10 @@ defmodule Honeybadger do
           end
         end
       _ ->
-        :ok
+        quote do
+          [_, _, _] = [unquote(exception), unquote(metadata), unquote(stacktrace)]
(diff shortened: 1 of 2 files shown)


nurugger07:
Awesome! Thanks :) I'll review it today

nurugger07:
Thanks! This looks good 👍

mitchellhenke:
Thank you! Glad I was able to help!

Issue #42, opened 2016-06-14
convert_time_unit maybe should use :native for FromUnits


KronicDeth:
There is an inconsistency with [`Plug.Logger`](https://github.com/elixir-lang/plug/blob/22227a5f1627b2bb0410eb4a0cfdc40f1b037c18/lib/plug/logger.ex#L47-L48):   It uses `:erlang.monotonic_time` to get the start and end times and when converting from the `:erlang.monotonic_time` default units to microseconds it does:

```
(stop - start) |> :erlang.convert_time_unit(:native, :micro_seconds)
```

[`HoneyBadger.Plug`](https://github.com/honeybadger-io/honeybadger-elixir/blob/master/lib/honeybadger/plug.ex#L12-L17) uses `:erlang.monotonic_time` to get the start and end times (the same as `Plug.Logger`), but when converting from the `:erlang.monotonic_time` default units it does:

```
response_time = :erlang.convert_time_unit(after_response - before_response, :nano_seconds, :milli_seconds)
```

So, `Plug.Logger` uses `:native` for `FromUnits` while `HoneyBadger.Plug` uses `:nano_seconds`.  They are probably equivalent on OSX and Linux, but I think it is probably safer for `HoneyBadger.Plug` to use `:native` like `Plug.Logger` in case it is used on a BEAM VM where `:nano_seconds` isn't the same as `:native`

nurugger07:
This sounds reasonable @KronicDeth and thanks for reporting it. It would make sense to have the two conform to the same conversion type. I'll get with @joshuap and verify the case for the original implementation but we should be able to make the change

nurugger07:
Thanks for the heads up @KronicDeth! Branch is merged :)

PR #49, opened 2016-07-21
fix convert unit bug

mitchellhenke:
Related to #48 and #42 

`:erlang.convert_time_unit` expects 3 arguments, so this should fix the issue.


nurugger07:
Awesome! Thanks :)

PR #50, opened 2016-07-27
reformat notice request payload

mitchellhenke:
Our Honeybadger reports were missing things like params, route, environment, etc. and based on http://docs.honeybadger.io/guides/api.html, it looks like it is because the keys held in the `plug_env` map should be at top level of the `request` key in the payload.

I tested it in our development environment, and it started to include more things (params, URL, browser) on the error page, but not longer afterwards, I noticed that on a different environment, this extra data started to show up on the error page as well, even though it was still using the the currently incorrect format.  Did something change on how the `plug_env` key is handled, and is this change still necessary?

### DIFF
--- a/lib/honeybadger/notice.ex
+++ b/lib/honeybadger/notice.ex
@@ -12,7 +12,9 @@ defmodule Honeybadger.Notice do
       tags: Dict.get(metadata, :tags, []),
       backtrace: backtrace
     }
-    request = Dict.take(metadata, [:plug_env, :context])
+    context = Dict.get(metadata, :context, %{})
+    request = Dict.get(metadata, :plug_env, %{})
+              |> Dict.put(:context, context)
 
     %__MODULE__{error: error,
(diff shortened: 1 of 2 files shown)


starrhorne:
Hi Mitchell,

Yep, we noticed the problem as well and implemented a fix on our back-end. We'll release a new version of the elixir library as soon as we can. Thanks for the PR!

starrhorne:
@nurugger07 - Hey Johnny, can you review this when you have a second?

nurugger07:
@starrhorne all ready on it :)

mitchellhenke:
ah cool, so we just happened to be working on fixing this at the same time? :)

starrhorne:
@mitchellhenke - yep :)
@nurugger07  - Thanks! You're fast!

nurugger07:
@mitchellhenke Thanks for contributing! I'll get a release out tonight :)

PR #51, opened 2016-08-09
Demonstrates repeated calls to API

cmwilhelm:
When using the Honeybadger plug, it seems that multiple attempts to call Honeybadger are being performed.

This PR serves to demonstrate the problem with a failing test case, by setting up a Bypass server standing in for Honeybadger API. This server appears to receive messages twice from the plug machinery.

(Most of the test code exists to inject new configuration after the compilation of the existing `PlugApp` so that `exclude_envs` is meaningly overridden.)

I was unable to reproduce this problem by calling `Honeybadger.notify` directly.

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -32,6 +32,7 @@ defmodule Honeybadger.Mixfile do
      {:ex_doc, "~> 0.7", only: :dev},
 
      # Test dependencies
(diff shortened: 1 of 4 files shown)


nurugger07:
@cmwilhelm this is resolved in master and will be available in the release on Monday (10/23)

Issue #57, opened 2016-09-13
Override Exception Name


zachdaniel:
When we do `Honeybadger.notify("Something went wrong")` how could we dynamically set the name of the exception that it reports to honey badger?

nurugger07:
@zachdaniel that PR was merged and will be available in the next release coming this week

PR #58, opened 2016-09-13
added the ability to send in a map with :class and :message keys for …

zachdaniel:
…dynamic error names/messages


zachdaniel:
This allows us to do #57, and is in fairly in line with the ruby client example.

```
  Honeybadger.notify(
    :error_class => "Special Error",
    :error_message => "Special Error: #{e.message}",
    :parameters => params
  )
```

PR #62, opened 2016-10-13
Add filter

pbm:
This PR adds a customizable filter to remove sensitive data before sending the notice to Honeybadger.io.

We needed some filtering on our app, so I went ahead and implemented something.  Not sure if this is the best approach, but it seems to be working for us.   I'm happy to adjust the design/features if desired.

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,20 +111,6 @@ rescue
 end
 ```
 
-### Filtering Sensitive Data
-
-Before data is sent to Honeybadger, it can be run through a filter to remove sensitive fields or do other processing on the data.  For basic filtering you can configure `Honeybadger.DefaultFilter` and set sensitive keys in `filter_keys`:
-
-```elixir
-config :honeybadger,
-  filter: Honeybadger.DefaultFilter,
-  filter_keys: [:password, :credit_card]
-```
-
-This will remove any entries in the context, session, cgi_data and params that match one of the filter keys.  The check is case insensitive and matches atoms or strings.
-
-If `Honeybadger.DefaultFilter` does not suit your needs, you can implement your own filter. See the `Honeybadger.Filter` module doc for details on implementing your own filter.
-
 ## Sample Application
 
 If you'd like to see the module in action before you integrate it with your apps, check out our [sample Phoenix application](https://github.com/honeybadger-io/crywolf-elixir).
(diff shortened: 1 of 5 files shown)


joshuap:
@pbm thanks! I like this. :+1:

Do you think it would make sense to provide a default filter which automatically filters keys like `:credit_card` and `:password` from the params? We do this for Rails.

@nurugger07 what do you think?

joshuap:
Maybe we should also add request filters for `:cgi_data` (which works the same as context and params), and `:url` (which would work like the error message string filter)?

pbm:
I'd be happy to add some sensible defaults. I'm without computer until Tuesday so won't be able to work on that until then.

nurugger07:
@joshuap I can add the sensible defaults to match the Rails defaults so that it can make the release for Monday

pbm:
@joshuap I'm back at a computer.  Do you want me to implement a default filter (perhaps along the lines of the ruby one, where it would take config params such as `filter_keys: [:password, :credit_card]` and `disable_session: true`),  or is @nurugger07 working on it?

joshuap:
@pbm I like your changes. @nurugger07 thoughts? My only comment is that I think we should use the `DefaultFilter` by default, and also default to `filter_keys: [:password, :credit_card]`. If people don't want the defaults they can change them, then. I'll leave this up to @nurugger07 to merge and make any other changes he wants for Monday's release! Let me know if either of you have questions. :metal:

pbm:
@joshuap I don't really like the DefaultFilter I put in, it is not a very clean abstraction.  I'd like to clearly separate the overridable functions from the mechanics of parsing/building notices.  I think I can have something later tonight.

pbm:
And I'll also make DefaultFilter default behavior with `[:credit_card, :password]`.

pbm:
@joshuap I'm done (I think...unless you tell me you need something else).  I made the default to filter credit card and password, and it is now easier to implement your own filter.  I also cleaned up the moduledocs etc. and added a few more tests.

nurugger07:
@pbm I'm reviewing this today. One thing to note, can you squash these commits into a single commit please? I'll get back to you on any other feedback so we can try to get this in Monday's release

pbm:
@nurugger07 Is the github  [squash-merge](https://github.com/blog/2141-squash-your-commits) not sufficient?  If I push -f a squashed version it can potentially mess with comments etc attached to code.  But I'll do it if squash-merge somehow doesn't meet your needs.

PR #69, opened 2016-11-18
Add ability to configure a HTTP proxy

cbarratt:
This is very rough and ready, I will test this today to ensure it works. I'm not the greatest Elixir dev so feel free to pick it apart!


joshuap:
@cbarratt thanks for the PR! I'm going to leave this open so that we can review it for a future version release. We are in the process of building a roadmap.

mwoods79:
@cbarratt can you pop the last commit? This should take care of the Poison version problems. https://github.com/honeybadger-io/honeybadger-elixir/pull/75

cbarratt:
@mwoods79 - I think that should be sorted now 👍

lib/honeybadger/client.ex
```diff
@@ -24,23 +33,25 @@ defmodule Honeybadger.Client do
...
-    )
+    post(client, @notices_endpoint, body, http_mod)
   end
 
-  defp headers(api_key) do
-    [{"Accept", "application/json"},
-     {"Content-Type", "application/json"},
-     {"X-API-Key", api_key}]
+  defp post(client, url, body, http_mod) do
+    proxy = [proxy: client.proxy, proxy_auth: client.proxy_auth]
```
mwoods79: Can you move this inside the case statement?

lib/honeybadger/client.ex
```diff
@@ -24,23 +33,25 @@ defmodule Honeybadger.Client do
...
+
+    case client.proxy do
+      nil -> http_mod.post(client.origin <> url, body, client.headers)
+      _ -> http_mod.post(client.origin <> url, body, client.headers, proxy)
+    end
   end
 
+  defp headers(api_key) do
+    @headers ++ [{"X-API-Key", api_key}]
```
mwoods79: Reversing the order of this will make it `O(1)` instead of `O(n)`. In this case `n=2`, but `@headers` may grow.

`[{"X-API-Key", api_key}] ++ @headers`

cbarratt:
@mwoods79 - No worries. I've just fixed the comments up and also merged in from master and noticed a possible issue when merging.

In master on line 29 of `send_notice` there was an assignment for `body` which was taking the notice map and encoding it to a binary for HTTPoison!

As you can see here https://github.com/honeybadger-io/honeybadger-elixir/commit/f4d2142fe898931352cc22f2287b80b35fb90eb1#diff-6b3f64819f88b8f95c5eeb8bc458970fR23

I have a feeling this will actually be breaking maybe, afaik HTTPoison only accepts a binary as the body

PR #71, opened 2017-01-19
Suppress metric HTTP calls in excluded environments

moxley:
In excluded environments, HTTP calls were suppressed for error notifications, but not for metrics reporting. This resulted in error logging noise during test runs, when ExVCR complained about un-mocked HTTP calls.

honeybadger-exvcr-noise


joshuap:
Thanks @moxley! I'll merge this and get a bugfix release out soon. In the next major version we'll also be removing metrics entirely since we no longer display them in the UI.

PR #80, opened 2017-03-08
Remove metrics

mwoods79:
Removing Metrics is more than just removing code. The metrics supervisor was started by the Application and now that it is gone the only thing that is required is the `plug`. The start function was also used to initialize application data.


mwoods79:
Not sure that the above rambling was clear. We can create a fake supervisor that does nothing (like the WIP @nurugger07 was working on). That will remove metrics but seems like a hack. Instead we should just remove the Application start function. In order to do this however, we need to rethink how the application config is loaded.

joshuap:
@mwoods79 [here's the PR which introduced metrics](https://github.com/honeybadger-io/honeybadger-elixir/pull/37/files). I thin we had the Application start function without a supervisor before metrics. Could we just go back to what we had before? Are there any new implications to consider?

mwoods79:
@joshuap I put it back the way it was before. This kicks the ball down the road some.

This part needs to be rewritten. I'm not even sure if we need an application with a start function. We may be able to simplify this to just a `plug`.

mwoods79:
@joshuap One other thing `Application.ensure_started(:httpoison)` returns `:ok` or `{:error, _term}`. This means when it fails it will return `{{:error, _term}, _pid}`. That breaks but is less than desirable for error handling.

PR #82, opened 2017-03-20
Add runtime config support for api_key

aditya7iyengar:
This would allow us to run the same builds on different machines without having to rebuild with a set of configs

### DIFF
--- a/lib/honeybadger.ex
+++ b/lib/honeybadger.ex
@@ -232,6 +232,9 @@ defmodule Honeybadger do
       filter_disable_session: false]
   end
 
+  defp resolve_api_key({:system, env_var}), do: System.get_env(env_var)
+  defp resolve_api_key(api_key), do: api_key
+
   defp require_environment_name! do
     if is_nil(Application.get_env(:honeybadger, :environment_name)) do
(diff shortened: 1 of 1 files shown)


mwoods79:
@aditya7iyengar Wouldn't this do the same thing?

```elixir
# config.exs
config :honeybadger,
  api_key: System.get_env("HONEYBADGER_API_KEY")
```

aditya7iyengar:
@mwoods79 It would, but only if we are creating a build with that config set up. Many companies build an application on a machine and deploy it to separate machines which could have different env variables. So, in this way they can bypass setting up configs at build-time.

Here's how phoenix does it:
https://github.com/phoenixframework/phoenix/blob/master/lib/phoenix/endpoint.ex#L157

aditya7iyengar:
bump @mwoods79

aditya7iyengar:
@mwoods79 Any updates on this one?

mwoods79:
@aditya7iyengar we are currently rethinking the way we initialize and store configuration data. Once we have a better solution (or even decide to keep the current one) then we can explore runtime config keys. Sorry you've been waiting so long for this feature, hopefully it won't be too much longer.

PR #86, opened 2017-05-03
Loosen restrictions on httpoison

mwoods79:
Allow for newer versions of httpoison. For some reason there was only one release of `0.10` and then a skip straight to `0.11`. I did not include `0.10`.

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -24,7 +24,7 @@ defmodule Honeybadger.Mixfile do
   end
 
   defp deps do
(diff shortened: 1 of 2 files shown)


drewblas:
Could this potentially just be `>= 0.9.0` instead?

mwoods79:
I would rather lock it down to known working versions. If `0.12` broke honeybadger-elixr then apps may not work properly, or report errors correctly.

PR #89, opened 2017-09-19
Add get_env/1 to universally support system variables

sorentwo:
This adds a `Honeybadger.get_env/1` function that universally fetches application or system variables via `:system` tuples. Resolution works as follows:

* If a value from `{:system, "ENV_NAME"}`  is missing it raises an `ArgumentError`, indicating the missing "ENV_NAME".
* If the requested env key is unknown it raises an `ArgumentError` stating that the key has not been set.

After adding the system tuple support I updated the few uses of `System.get_env`. This removes any reliance on the system env during compile time, which simplifies configuring releases or deploying to Heroku.

### DIFF
--- a/lib/honeybadger.ex
+++ b/lib/honeybadger.ex
@@ -199,6 +199,11 @@ defmodule Honeybadger do
 
   @doc """
   Fetch configuration specific to the :honeybadger application.
+
+  # Example
+
+      Honeybadger.get_env(:exclude_envs)
+      #=> [:dev, :test]
   """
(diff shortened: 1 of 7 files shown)


joshuap:
@sorentwo thanks for these PRs; just wanted to let you know they are on my radar. Will take a closer look soon!

sorentwo:
@joshuap Thanks for letting me know. A few more are on the way as well!

sorentwo:
@joshuap I'm going to close this in favor of #93, which is built off of this.

PR #91, opened 2017-09-19
Update the CI build matrix

sorentwo:
The matrix lacked a 1.5 release. Additionally, all of the versions within the matrix are compatible with OTP 19, which is likely to be the oldest OTP most applications are running on now.


joshuap:
👍 thanks!

Issue #95, opened 2017-10-10
Tagged releases are behind


joshuap:
We haven't been tagging our releases properly: https://github.com/honeybadger-io/honeybadger-elixir/releases

We should add tags for all the versions on Hex:

https://hex.pm/packages/honeybadger

This will require digging through the commit history for the release shas, unfortunately.

Moving forward, let's always tag our stable releases (alpha/beta/rc releases do not require tags).

- [x] Add tags for historical releases on Hex
- [x] Add a `release` task that automates releasing a new version, similar to honeybadger-ruby.
- [x] Document

minhajuddin:
#98 Adds the release task and the documentation

For the first todo list item, hex does not store the git sha id, so there is no definite way to know which commit. The following script gets the commits and versions whenever the version property was updated. Hopefully, that would be the commit that created the hex release.

```
honeybadger-elixir [hex-publish] $ for f in $(git log --oneline mix.exs | cut -f1 -d' '); do echo $f $(git show $f -- mix.exs | grep -E '^\+' | grep version) ; done | grep version
f4a49dc + version: "0.7.0-beta",
db801bd + version: "0.6.3",
73c896c + version: "0.6.2",
3468a90 + version: "0.6.1",
355bc8e + version: "0.6.0",
d17b0bc + version: "0.5.0",
9d81f4e + version: "0.4.0",
bfe2acf + version: "0.3.1",
578566b + version: "0.3.0",
90e0c73 + version: "0.2.0",
6c5e31d + version: "0.1.2",
9672533 + version: "0.1.0",
b7d55c1 + version: "0.0.1",
honeybadger-elixir [hex-publish] $ 
```

You can run the following command to create the proper tags for all our hex releases

```bash
git tag --annotate --message "tag for hex.pm release" "0.6.3" db801bd
git tag --annotate --message "tag for hex.pm release" "0.6.2" 73c896c
git tag --annotate --message "tag for hex.pm release" "0.6.1" 3468a90
git tag --annotate --message "tag for hex.pm release" "0.6.0" 355bc8e
git tag --annotate --message "tag for hex.pm release" "0.5.0" d17b0bc
git tag --annotate --message "tag for hex.pm release" "0.4.0" 9d81f4e
git tag --annotate --message "tag for hex.pm release" "0.3.1" bfe2acf
git tag --annotate --message "tag for hex.pm release" "0.3.0" 578566b
[...]

PR #98, opened 2017-10-14
Add a mix task for publishing on hex and creating git tags

minhajuddin:
This adds a `hex_publish` mix task, which does the things mentioned on #95


joshuap:
Looks great, thanks @minhajuddin

Issue #92, opened 2017-09-21
Support some extra `Honeybadger.notify` signatures


joshuap:
It would be nice to do the following (which we also have in honeybadger-ruby):

```elixir
Honeybadger.notify(%{:error_message => "the message", :error_class => "CustomError"})
Honeybadger.notify("the message")
```

This isn't an exhaustive list.

You can also currently do `Honeybadger.notify(%RuntimeError{message: "the message"})` (according to the module doc), but it isn't documented in the README.

PR #104, opened 2017-10-18
Add explicit support for sending string noticies

sorentwo:
While it was possible to send an arbitrary notice by constructing an error manually, it wasn't possible to pass a string to the `notify` function. Sending a notice directly from a string is more convenient, and it make the exact arguments that the `notify` function accepts more obvious.

This refactors the `Honeybadger.Notice` module a little to guard the accepted types a bit more. Part of the guarding process involved adding typespecs for the components that make up a `Honeybadger.Notice` struct.

Addresses #92

@joshuap This is a little yak shaving to prepare `Honeybadger.notify/3` to handle arbitrary error logging with iodata.


lib/honeybadger.ex
```diff
@@ -174,13 +174,25 @@ defmodule Honeybadger do
...
-      Honeybadger.notify(%MyException{}, %{culprit_id: 123})
-      #=> :ok
+      iex> Honeybadger.notify(%RuntimeError{}, %{culprit_id: 123})
+      :ok
   """
-  @spec notify(Exception.t, Map.t, list()) :: :ok
-  def notify(exception, metadata \\ %{}, stacktrace \\ nil) do
+  @spec notify(Notice.noticeable, Map.t, list) :: :ok
+  def notify(exception, metadata \\ %{}, stacktrace \\ nil)
+      when is_map(exception) or is_binary(exception) do
```
joshuap: I was a little confused by the line break/indentation of the guard clause here. Is this format standard?

lib/honeybadger/notice.ex
```diff
@@ -1,43 +1,76 @@
...
 
-  def new(error, metadata \\ %{}, backtrace)
+      iex> Honeybadger.Notice.new(:badarg, %{}, []).error
+      %{backtrace: [], class: "ArgumentError", message: "argument error", tags: []}
+
+      iex> Honeybadger.Notice.new(%RuntimeError{message: "oops"}, %{}, []).error
+      %{backtrace: [], class: "RuntimeError", message: "oops", tags: []}
+  """
+  @spec new(noticeable, Map.t, list) :: t
+  def new(error, metadata, backtrace)
```
joshuap: Do you mind explaining how this line works? I can't remember what Elixir does when there is no function body.

joshuap:
@sorentwo btw, any idea why the build is failing?

sorentwo:
@joshuap I took a look at the run time of that job and it is a full 2-3 seconds longer than others. The failures were from message assertion timeouts, so I think things were running slowly. It is a shared environment, so possibly some noisy neighbors. Re-running the build should fix it.

In a related note, sorry I didn't notice this before. I don't get any notifications from Travis when the build breaks.

lib/honeybadger.ex
sorentwo: That is the standard format and the way the new built in formatter will indent guard clauses that exceed the maximum line width.

lib/honeybadger/notice.ex
sorentwo: It defines a function head without any body. Usually you do this to specify defaults for functions with multiple clauses. In this case I did it for the `@spec` above it. Functions can only have one spec, regardless of how many clauses there are.

joshuap:
No worries. :) I tried assigning you to the PR, not sure if that will make you get notifications.

Hrmm, I'm not sure why but that build keeps failing (2 attempts so far). I can wait a while and try again.

sorentwo:
Thanks for trying the rebuild. I tried it locally and got the same error. It turns out that the Erlang version of the argument error is simply `:badarg`, which is an atom. The `Notice.new` function allowed it, but `Honeybadger.notify` didn't.

I've pushed up a fix (and an explicit test). Glad the tests caught it.

lib/honeybadger/notice.ex
```diff
@@ -1,43 +1,76 @@
...
-                notifier: notifier(),
-                server: server()}
-     |> filter(Honeybadger.get_env(:notice_filter))
+    request =
+      metadata
+      |> Map.get(:plug_env, %{})
+      |> Map.put(:context, context)
+
+    __MODULE__
+    |> struct!(error: error, request: request, notifier: notifier(), server: server())
```
minhajuddin: Using `%Notice{error: ... }` may be more readable than using `struct!`

lib/honeybadger.ex
```diff
@@ -174,13 +174,30 @@ defmodule Honeybadger do
...
-      Honeybadger.notify(%MyException{}, %{culprit_id: 123})
-      #=> :ok
+      iex> Honeybadger.notify(%RuntimeError{}, %{culprit_id: 123})
+      :ok
   """
-  @spec notify(Exception.t, Map.t, list()) :: :ok
-  def notify(exception, metadata \\ %{}, stacktrace \\ nil) do
+  @spec notify(Notice.noticeable, Map.t, list) :: :ok
+  def notify(exception, metadata \\ %{}, stacktrace \\ nil)
+      when is_map(exception) or is_binary(exception) or is_atom(exception) do
```
minhajuddin: The `notify` function allowed all erlang terms before this change. If we restrict this to just maps, binaries and atoms, we may inadvertently break other's code.

As of now, when you pass a string or iodata or integer or any erlang term, the call gets forwarded to the `Notice.new` function at https://github.com/honeybadger-io/honeybadger-elixir/blob/master/lib/honeybadger/notice.ex#L19 . This further calls the `exception = Exception.normalize(:error, exception)` function which allows the `exception` variable to have any kind of erlang term.

Explicitly documenting the fact that we accept a string is a good thing, we should probably drop the guard clause.

sorentwo: Excellent point! The guard clause has been removed.

lib/honeybadger/notice.ex
sorentwo: Using `%Notice{}` is definitely more readable. I was aiming to leverage `@enforce_keys`, but that is checked at compile time for literal constructors anyhow. This has been changed (with an `alias __MODULE__` at the top to avoid writing `%__MODULE__{}`

Issue #36, opened 2016-02-11
Add a mix task to test reporting an exception from a phoenix controller


joshuap:
Not sure if this is possible (or easy), but it would be cool to do this for Phoenix similar to how we do for Rails.

rbishop:
This should be possible! Would you want the mix task in this app or in a phoenix? I'll have to take a look at and seeing how you do it for Ruby/Rails.

PR #107, opened 2017-10-26
Honeybadger test mix task

minhajuddin:
Adds the ability to simulate an error using a mix task. Completes #36


joshuap:
@minhajuddin will this report a live error in the development environment by default, or do you need to add the `exclude_envs` config first? If the latter, is it possible to make this task always report a live error?

minhajuddin:
@joshuap Good point about the `excluded_envs`. I have added some code which will mute them before running the task.

joshuap:
Looks good, thanks! One last thing, can you add a changelog entry? Feel free to squash ans merge, the.

minhajuddin:
Added changelog entry and squashed the commits.

PR #109, opened 2017-10-30
Re-implement error logger using gen_event behaviour

sorentwo:
* Implement `Honeybadger.stop/1` callback to de-register the error logger when the supervision tree is shut down.
* Prevent registering duplicate loggers in tests, this interfered with correctly testing api responses.
* Remove use of `GenEvent`, replacing it with direct use of the `gen_event` behaviour. No more deprecation warnings!
* Add more detailed assertions in the logger tests, now the error "class" and the context are asserted.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ adheres to [Semantic Versioning](http://semver.org/).
   possible to send a `RuntimeError` by calling `Honeybadger.notify("oops!")`.
 
 ### Changed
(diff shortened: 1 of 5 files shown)


joshuap:
@sorentwo could you add an entry to the CHANGELOG? I keep forgetting about that.

sorentwo:
@joshuap Changelog has been updated, thanks for the reminder. 👍

minhajuddin:
Thanks for the PR @sorentwo I'll spend some time tomorrow to review this :+1:

PR #114, opened 2017-11-12
Safely handle unknown environment values

sorentwo:
The use of `String.to_existing_atom/1` would cause the client to crash if an unknown binary was provided. For example, on a staging server using a `{:system, "HONEYBADGER_ENV"}` tuple, the app would fail to boot.

This replaces `to_existing_atom/1` with `to_atom`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Fixed
+- Safely convert binary `:environment_name` values to an atom. If the
+  environment was specified via `{:system, "HONEYBADGER_ENV"}` and the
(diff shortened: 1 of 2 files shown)


minhajuddin:
@sorentwo Would there be a valid scenario when a valid environment name atom wouldn't be present in the BEAM? As far as I can think this would happen only in an instance where there is a typo in the environment name in which case crashing should be ok.

sorentwo:
This is currently happening on one of our staging servers, which is how I discovered the issue. The server uses `prod` for everything, only honeybadger is configured to use `staging`. There really isn’t use of the atom `staging` anywhere else.

maxim:
Would appreciate getting this fix released, we are having the same issue.

joshuap:
@maxim done! https://hex.pm/packages/honeybadger/0.8.0

Issue #78, opened 2017-03-01
Track parameters from function invocations


joshuap:
In some cases it appears that a backtrace can include the parameters used to invoke a function that caused an error. Example:

```
2017-03-01T16:53:52.772605+00:00 app[web.1]: (my_web) web/controllers/gateway_controller.ex:126: MyWeb.GatewayController.my_function("", nil, 608, 2) 
```

It would be cool if we could capture these and display them in Honeybadger.

duff:
👍 Thanks Josh!

sorentwo:
@joshuap I did a little research spike on this and it looks rather straight forward. Do you have a preferred/suggested format for reporting arguments? Looking at the Ruby gem's implementation didn't provide any insight.

joshuap:
@sorentwo for reference, [here is our collector API documentation](http://docs.honeybadger.io/guides/api.html#collector-api). I'm thinking that we'd add some new property to the backtrace, such as  `.error.backtrace[].params`:

```
{
  "error":{
    "class":"RuntimeError",

    # ...

    "backtrace":[
      {
        "number":"4",
        "file":"web/controllers/gateway_controller.ex",
        "method":"my_function",
        "params": ["", nil, 608, 2]
      },

      # ..

    ]

    # ...

}
```

What do you think?

joshuap:
cc @stympy -- thoughts on this?

sorentwo:
@joshuap Adding a new entry for `params` sounds good to me, provided long as the server will safely accept those.

joshuap:
Agreed. We can add support on the server; my only concern is leaking sensitive data. This should probably be an optional feature, or maybe integrate with the request filters somehow.

sorentwo:
An argument could be anything, even a very sizable struct like a `Plug.Conn`, which itself could have contain sensitive data. Additionally, there isn't any way to be sure what a particular argument is, so it could be a name and password, but without any type of label.

I think it should be an opt-in feature with caveats about sensitive information.

This is definitely something I would opt into for the primary app I run, as it doesn't handle any sensitive info (passwords, payment) but knowing the args passed would be very helpful.

stympy:
This looks good to me.  Is params always going to be an array, even when just one arg is passed to the function?  Consider this question a 👍 for it always being an array. 😄

PR #115, opened 2017-11-13
Include params and arity in formatted backtrace

sorentwo:
This enhances the backtrace to include function arity and arguments, if available. The arguments are inspected to ensure they can be serialized as JSON without any issues.

Argument inclusion is configured with `filter_args`, which defaults to `false`. Therefore by default arguments are included in a notice's backtrace. When arguments are filtered an empty list will be returned instead.

Also note, Poison can't encode tuples, structs, pids, refs, etc. so `inspect` is called on each argument. It may be preferable to be more selective with argument formatting, as a struct or list may be quite large. Any thoughts on this?

Addresses #78

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ config :honeybadger,
 If `environment_name` is not set we will fall back to the value of `Mix.env()`.
 `Mix.env()` uses the atomized value of the `MIX_ENV` environment variable and
 defaults to `:dev` when the environment variable is not set. This should be good
-for most setups. If you want to have an environment_name which is different than
+for most setups. If you want to have an `environment_name` which is different than
 the `Mix.env()`, you should set `environment_name` in your `config.exs` files for each
 environment. This ensures that we can give you accurate environment information
 even during compile time. Explicitly setting the `environment_name` config
@@ -140,7 +140,7 @@ config :honeybadger,
   filter_keys: [:password, :credit_card]
 ```
 
-This will remove any entries in the context, session, cgi_data and params that match one of the filter keys.  The filter is case insensitive and matches atoms or strings.
+This will remove any entries in the `context`, `session`, `cgi_data` and `params` that match one of the filter keys.  The filter is case insensitive and matches atoms or strings.
 
 If `Honeybadger.Filter.Default` does not suit your needs, you can implement your own filter. See the `Honeybadger.Filter.Mixin` module doc for details on implementing your own filter.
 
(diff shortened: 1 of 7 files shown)


lib/honeybadger/backtrace.ex
```diff
@@ -1,33 +1,53 @@
...
+    format_method(fun, length(args))
+  end
+  defp format_method(fun, arity) when is_integer(arity) do
+    "#{fun}/#{arity}"
+  end
+
+  defp format_args(args, false = _filter) when is_list(args) do
+    Enum.map(args, &inspect/1)
+  end
+  defp format_args(_arity, _filter) do
```
minhajuddin: Can we rename this to `_args`

mix.exs
```diff
@@ -36,6 +36,7 @@ defmodule Honeybadger.Mixfile do
            notice_filter: Honeybadger.NoticeFilter.Default,
            filter: Honeybadger.Filter.Default,
            filter_keys: [:password, :credit_card],
+           filter_args: false,
```
minhajuddin: If this is an opt-in feature, we should default to `filter_args: true`. @joshuap thoughts?

joshuap: Yeah, I'm thinking we want to disable this feature by default. We discussed making it opt-in in https://github.com/honeybadger-io/honeybadger-elixir/issues/78#issuecomment-342577068. Does that make sense, @sorentwo?

joshuap:
> Also note, Poison can't encode tuples, structs, pids, refs, etc. so inspect is called on each argument. It may be preferable to be more selective with argument formatting, as a struct or list may be quite large. Any thoughts on this?

I think we *do* want to do something to handle super large structures. In our ruby gem we truncate the maximum string size that can be sent as any value to 64KiB. That's more of a catch-all which applies to any data (including context). We also apply some extra sanitization to `#inspect` output by default -- for instance, `#<MyObject id=1 foo=bar ... >` becomes `#<MyObject>`. We *could* do something similar here, although I'm not sure what makes sense yet. The easiest thing to do would probably be to come up with some standard representation of un-encodable objects.

mix.exs
sorentwo: I've reversed the default, it is now `true`, making it opt-in like we discussed 👍

lib/honeybadger/backtrace.ex
sorentwo: When it isn't a list it is a number, which really is the arity of the function. Stack traces are guaranteed to have _either_ the args or arity, so this is indicating what the argument really is.

README.md
```diff
@@ -157,22 +157,21 @@ config :honeybadger,
 
 Here are all of the options you can pass in the keyword list:
 
-| Name         | Description                                                               | Default |
-|--------------|---------------------------------------------------------------------------|---------|
-| api_key      | Your application's Honeybadger API key                                    | System.get_env("HONEYBADGER_API_KEY"))` |
-| environment_name | (required) The name of the environment your app is running in.                   | nil |
-| app          | Name of your app's OTP Application as an atom                             | Mix.Project.config[:app] |
-| use_logger   | Enable the Honeybadger Logger for handling errors outside of web requests | true |
```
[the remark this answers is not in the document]
sorentwo: You're correct, I did! I've restored it.

sorentwo:
@joshuap I've started on a spike to implement customized `to_string` handling with a protocol. It will provide conservative defaults for all Elixir types and we or other users can add new implementations to extend it if needed.

Going with `%MyObject{}` over `%MyObject{id: 1, foo: :bar}` makes a lot of sense to me, so I've started that direction.

lib/honeybadger/backtrace.ex
minhajuddin: My bad :) I thought the first argument would always be a list. I thought, this head was to match when filter was false. 👍

minhajuddin:
@sorentwo Is there anything pending on this PR? Are you planning on pushing a new commit with the review suggestion changes?

sorentwo:
Yes, there is work pending and a commit that I haven’t pushed. I’ve been on vacation for the past week, this will be a little while longer.

sorentwo:
@minhajuddin @joshuap The changes to inspected arguments have been pushed up. We're using `Kernel.inspect/2` with some custom options. This handles every possible structure, and with `limit` and `printable_limit` we get protection against large lists, structs, and binaries.

I've set the list/tuple/struct limit to 5 elements, and the printed limit to 1024 bytes. Those felt like reasonable amounts, let me know what you think. The values are all configured in a module attribute at the top of `Honeybadger.Backtrace`, and would be easy to modify.

minhajuddin:
Looks great :+1:

sorentwo:
Is anything blocking this still? I'd love to get it merged, it would be helpful with a few errors I'm seeing currently.

joshuap:
@sorentwo we still need to implement on the backend, sorry for the delay. I'll look into that next.

joshuap:
@sorentwo I'm trying to test this out in my Phoenix app. I'm getting the function arity, but I'm still getting empty args when I set the config `filter_args: false` in my phoenix app. Any ideas what I'm doing wrong?

```
 elixir --version
Erlang/OTP 20 [erts-9.1.5] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:10] [hipe] [kernel-poll:false] [dtrace]

Elixir 1.5.2
```

sorentwo:
@joshuap According to this [post on erlang questions](http://erlang.2086793.n4.nabble.com/getting-args-with-erlang-get-stacktrace-or-similar-td4720098.html) by Lukas Larsson (@garazdawi) the only time it will ever give arguments instead of arity is:

> You get that information if the error is a function clause, however if you crash for any other reason (badmatch, badarith, etc) within a function you cannot get that information.
>
> The reason for this is because the compiler may have found that the arguments to the function in which you crashed to be dead and optimized the values away.

I've been experimenting with causing various errors and checking the stacktrace, I only seem to be able to get the arity. I'm not sure what else to do here, I'll try looking at how elixir/iex manages to display arguments and clauses so thoroughly.

sorentwo:
Very relevant to this discussion, and the future handling of stacktraces: https://github.com/elixir-lang/elixir/issues/7097

joshuap:
I created issue #123 to track improvements to argument handling. In any case, Honeybadger now supports displaying arguments with the stack trace, so I'm going to merge this!

PR #117, opened 2017-11-15
Deactivate error tracking and reporting when in exclude_envs

### LINKED ISSUE #116: Unable to run application tests when HONEYBADGER_API_KEY is not set
We just updated from 0.6.3 to 0.7.0, and discovered that we are no longer able to run our test suite without setting `HONEYBADGER_API_KEY` in the environment. This doesn't really make sense, because the test environment is listed in `:excluded_envs` in the honeybadger config.

Our quick workaround is to add an empty `HONEYBADGER_API_KEY` to our env, but it would be better if we only had to export the variable for environments where honeybadger was actually enabled.

```
** (Mix) Could not start application honeybadger: exited in: Honeybadger.start(:normal, [])
    ** (EXIT) an exception was raised:
        ** (ArgumentError) system variable "HONEYBADGER_API_KEY" is not set
            (honeybadger) lib/honeybadger.ex:233: Honeybadger.get_env/1
            (honeybadger) lib/honeybadger.ex:255: anonymous fn/2 in Honeybadger.get_all_env/0
            (elixir) lib/enum.ex:1811: Enum."-reduce/3-lists^foldl/2-0-"/3
            (honeybadger) lib/honeybadger.ex:254: Honeybadger.get_all_env/0
            (honeybadger) lib/honeybadger.ex:146: Honeybadger.start/2
            (kernel) application_master.erl:273: :application_master.start_it_old/4
```

minhajuddin:
Closes #116 

WIP, don't merge yet.

### DIFF
--- a/config/test.exs
+++ b/config/test.exs
@@ -3,7 +3,8 @@ use Mix.Config
 config :honeybadger,
   environment_name: :test,
   api_key: "abc123",
-  origin: "http://localhost:4444"
+  origin: "http://localhost:4444",
+  exclude_envs: []
 
 config :ex_unit,
   assert_receive_timeout: 400,
--- a/lib/honeybadger.ex
+++ b/lib/honeybadger.ex
@@ -140,6 +140,16 @@ defmodule Honeybadger do
 
   @doc false
   def start(_type, _opts) do
+    children = if Honeybadger.Utils.enabled?() do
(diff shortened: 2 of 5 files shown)


minhajuddin:
This PR doesn't completely fix #116 . Created #118 to fix it.

PR #118, opened 2017-11-18
Fix failing tests in host apps

### LINKED ISSUE #116: Unable to run application tests when HONEYBADGER_API_KEY is not set

minhajuddin:
Fixes #116

### DIFF
--- a/lib/honeybadger/client.ex
+++ b/lib/honeybadger/client.ex
@@ -72,6 +72,7 @@ defmodule Honeybadger.Client do
 
   def init(state) do
     warn_if_incomplete_env(state)
+    warn_in_dev_mode(state)
     :ok = :hackney_pool.start_pool(__MODULE__, [max_connections: @max_connections])
 
     {:ok, state}
@@ -87,6 +88,14 @@ defmodule Honeybadger.Client do
     end)
   end
   defp warn_if_incomplete_env(_), do: :ok
+
+  defp warn_in_dev_mode(%{enabled: false}) do
+    Logger.warn(
+      "Development mode is enabled. Data will not be reported until you deploy your app."
+    )
+  end
+
+  defp warn_in_dev_mode(_), do: :ok
 
   def terminate(_reason, _state) do
     :ok = :hackney_pool.stop_pool(__MODULE__)
--- a/lib/honeybadger/client.ex
+++ b/lib/honeybadger/client.ex
@@ -15,7 +15,7 @@ defmodule Honeybadger.Client do
 
   # State
 
-  defstruct [:enabled, :headers, :proxy, :proxy_auth, :url]
+  defstruct [:api_key, :enabled, :headers, :proxy, :proxy_auth, :url]
 
   # API
 
@@ -28,6 +28,7 @@ defmodule Honeybadger.Client do
   @doc false
   def new(opts) do
     %__MODULE__{enabled: enabled?(opts),
+                api_key: get_env(opts, :api_key),
                 headers: build_headers(opts),
                 proxy: get_env(opts, :proxy),
                 proxy_auth: get_env(opts, :proxy_auth),
@@ -102,6 +103,10 @@ defmodule Honeybadger.Client do
   end
 
   def handle_cast({:notice, _notice}, %{enabled: false} = state) do
+    {:noreply, state}
+  end
+
+  def handle_cast({:notice, _notice}, %{api_key: nil} = state) do
     {:noreply, state}
   end
 
--- a/lib/honeybadger/client.ex
+++ b/lib/honeybadger/client.ex
@@ -71,10 +71,22 @@ defmodule Honeybadger.Client do
   # Callbacks
 
   def init(state) do
+    warn_if_incomplete_env(state)
     :ok = :hackney_pool.start_pool(__MODULE__, [max_connections: @max_connections])
 
     {:ok, state}
   end
+
+  @mandatory_keys ~w[api_key environment_name]a
+  defp warn_if_incomplete_env(%{enabled: true}) do
+    @mandatory_keys
+    |> Enum.each(fn key ->
+      if !Honeybadger.get_env(key) do
+        Logger.error("mandatory :honeybadger config key #{key} not set")
+      end
+    end)
+  end
+  defp warn_if_incomplete_env(_), do: :ok
 
   def terminate(_reason, _state) do
     :ok = :hackney_pool.stop_pool(__MODULE__)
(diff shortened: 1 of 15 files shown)


joshuap:
@minhajuddin this also changes the behavior in non-dev environments; for example, the app will now boot with no API key in production. I'm not totally *against* allowing that, but I want to make sure we're explicitly making that decision before merging this. A few questions:

1. Do we want to take a softer approach to missing configuration options? The benefit of raising when config is missing is that it alerts the user immediately that errors won't be reported. The drawback is that it could crash their app if they forget to add the API key. I'm not sure which is most preferred in Elixir.
2. What happens when an API key isn't configure in a non-dev environment? Are error reports sent without an API key? In our Ruby gem we just log that the gem isn't configured instead of reporting the error.

/cc @sorentwo -- any thoughts on this?

minhajuddin:
> What happens when an API key isn't configure in a non-dev environment? Are error reports sent without an API key? In our Ruby gem we just log that the gem isn't configured instead of reporting the error.

In an excluded_env, we create our payload but skip sending it to honeybadger. This is not the optimal thing.

We have three surface points that need to change based on our env.

 1. Starting a GenServer for the client: We should ideally not start a GenServer when we are in an excluded_env. However, we do a bit of initialization of our configuration in the GenServer. This should probably be moved to some other place.
 2. `Honeybadger.notify`: This builds up a context even in excluded envs and fails if honeybadger configuration hasn't been fully setup. Ideally, we should replace this function with a noop when in an excluded env
 3. `Honeybadger.Plug`: This too builds up a lot of context and calls the notify. Ideally this plug should be a noop too in an excluded env.

I took a stab at these but ran into a few issues with configuration not being available at compile time. Let me take another stab at this tomorrow. Ideally we should let honeybadger crash the app for lack of an api_key as that will make people either add the env to `exclude_env` or set up an `api_key`

sorentwo:
I err on the side of logging for missing values rather than enforcing them with a runtime error. With the new mix task to verify setup users can be confident that they got things working.

Elixir already has facilities to make sure applications aren't started unnecessarily, either by specifying them as only loading in certain environments or excluding them from the applications list in older versions.

In any case, sprinkling multiple modules with extra conditionals seems heavy handed and prone to breakage.

minhajuddin:
@joshuap I think with the way the app is structured, this is the minimal patch that will fix the error and also show a warning if honeybadger is started in a non excluded env without an api_key. Please review this PR and let me know if you'd like to change anything.

joshuap:
I think this needs a bit more work/thought. There are really two cases here:

1. When in an excluded environment. In this case we don't care about valid configuration; we just want to warn the user that reporting live data to Honeybadger is disabled. In our ruby gem we log this message:

> Development mode is enabled. Data will not be reported until you deploy your app.

I'm OK with still going through some steps to build the payload, etc. in an excluded environment as long as the notice is not actually sent. Stubbing out the server logic late in the process has the benefit of covering more honeybadger code in the user's tests.

2. When in a live environment, we want to warn the user when an API key isn't present. Honestly, the API key is the only thing that we really need -- our API/UI can handle null environments. So maybe we should just log a warning if the API key isn't configured.

We also should make sure that if an error is reported *and there is no valid API key*, we don't attempt to send the notice.

@minhajuddin what do you think?

minhajuddin:
> When in an excluded environment. In this case we don't care about valid configuration...
We now show a warning in this scenario.

> We also should make sure that if an error is reported and there is no valid API key, we don't attempt to send the notice.
We now don't send notices when there is no valid api key

lib/honeybadger/client.ex
```diff
@@ -71,11 +72,32 @@ defmodule Honeybadger.Client do
   # Callbacks
 
   def init(state) do
+    warn_if_incomplete_env(state)
```
joshuap: What currently happens if an API key is configured but the environment name isn't? I'm wondering if we still need to make `environment_name` mandatory or if this could just warn if the API key is missing.

lib/honeybadger/client.ex
```diff
@@ -71,11 +72,32 @@ defmodule Honeybadger.Client do
...
 
     {:ok, state}
   end
 
+  @mandatory_keys ~w[api_key environment_name]a
+  defp warn_if_incomplete_env(%{enabled: true}) do
+    @mandatory_keys
+    |> Enum.each(fn key ->
+      if !Honeybadger.get_env(key) do
+        Logger.error("mandatory :honeybadger config key #{key} not set")
```
joshuap: I think this should be `Logger.warn`, like the dev mode warning. This isn't really an error.

lib/honeybadger/client.ex
minhajuddin: > What currently happens if an API key is configured but the environment name isn't?

It prints out an error `mandatory :honeybadger config key :environment_name not set`. The default value of `:environment_name` is set to `Mix.env()`, so it wouldn't be nil unless it is set to nil (https://github.com/honeybadger-io/honeybadger-elixir/blob/master/mix.exs#L30).

However, it doesn't stop sending notices. It stops sending notices only if the `api_key` is missing.

lib/honeybadger/client.ex
minhajuddin: This runs only in a non excluded env. Shouldn't we print an error if `api_key` is missing? This however checks for both `api_key` and `environment_name`

joshuap: Ah yeah, that sounds right.

lib/honeybadger/client.ex
joshuap: Makes sense. Let's just roll with this.

PR #120, opened 2017-12-05
Allow handle_errors to be overridden

aaronrenner:
I've been looking for a way to skip reporting certain types of errors and I found it would be incredibly helpful if I could override the `handle_errors/2` function that `Honeybadger.Plug` generates. This change would allow me to do this:

```elixir
defmodule MyAppWeb.Router
  use MyAppWeb, :router
  use Honeybadger.Plug

  # Don't report UnauthenticatedError to Honeybadger
  def handle_errors(_conn, %{reason: MyAppWeb.UnauthenticatedError{}}), do: nil
  def handle_errors(conn, error_struct), do: super(conn, error_struct)
end
```
If there's a better way to do this, I'd be happy to hear it. I'd just prefer not to report all of my `UnauthenticatedError`s to Honeybadger.

### DIFF
--- a/lib/honeybadger/plug.ex
+++ b/lib/honeybadger/plug.ex
@@ -9,22 +9,24 @@ defmodule Honeybadger.Plug do
       @phoenix Keyword.get(unquote(opts), :phoenix, :code.is_loaded(Phoenix))
 
       # Exceptions raised on non-existent Plug routes are ignored
-      defp handle_errors(conn, %{reason: %FunctionClauseError{function: :do_match}} = ex) do
+      def handle_errors(conn, %{reason: %FunctionClauseError{function: :do_match}} = ex) do
         nil
       end
 
       if @phoenix do
         # Exceptions raised on non-existent Phoenix routes are ignored
-        defp handle_errors(conn, %{reason: %Phoenix.Router.NoRouteError{}} = ex) do
+        def handle_errors(conn, %{reason: %Phoenix.Router.NoRouteError{}} = ex) do
           nil
         end
       end
 
-      defp handle_errors(conn, %{kind: _kind, reason: exception, stack: stack}) do
+      def handle_errors(conn, %{kind: _kind, reason: exception, stack: stack}) do
         metadata = %{plug_env: build_plug_env(conn, __MODULE__, @phoenix),
                      context: Honeybadger.context()}
         Honeybadger.notify(exception, metadata, stack)
       end
+
+      defoverridable [handle_errors: 2]
     end
   end
 


sorentwo:
I think `Honeybadger.Plug` should be more extensible all around, so this is a good start. That said, ignoring errors probably shouldn't require overriding a function. Users should be able to configure an additional list of errors to ignore, much like is possible with the ruby library.

Thoughts @joshuap?

minhajuddin:
@aaronrenner I definitely see that allowing users to ignore plug errors is useful. However, like @sorentwo mentioned it should be easier than overriding the `handle_errors` function. Can you change the patch so that it takes a list of struct matches of errors which can be ignored. Something like below:

```elixir
defmodule MyAppWeb.Router
  use MyAppWeb, :router
  use Honeybadger.Plug, ignore: [%MyAppWeb.UnauthenticatedError{}, MyAppWeb.AuhorizationFailure{}, ...]

  # change our `Honeybadger.Plug` to generate these ignore function heads 

end
```

sorentwo:
Doing it as an option passed to the use macro reads nicely, but it won’t help with async tasks or background jobs. I think it should be something configured for the entire application.

joshuap:
I tend to agree with @sorentwo, with caveats.

In our Ruby gem, we have the ignore config option, which is used like this:

```ruby
Honeybadger.configure do |config|
  config.exceptions.ignore += [UnauthenticatedError, AuhorizationFailure] # The += adds to an existing list which has a few sensible defaults for Rails, such  as ActiveRecord::RecordNotFound
end
```

This works for most people and is an extremely useful option.

One argument *for* having a Plug-specific (or context-specific) option -- possibly in addition to a global option -- is that some errors should only be ignored in a certain context, such as when they occur in a web request cycle.

We've run into an issue in Ruby where ignoring `ActiveRecord::RecordNotFound` makes sense in a Rails app because Rails automatically handles presenting that error to the user (as a 404 response) and is therefore non-exceptional. However, you may still want to report `ActiveRecord::RecordNotFound` in a background job (like in Sidekiq), as it's not treated as a 404.

One solution to this (when using global configuration) is to have different configurations for your web server process vs. your background worker process, but:

- This may not apply as well to Elixir since everything tends to run in a single process.
- It doesn't solve the issue of having a global default, which we do for `ActiveRecord::RecordNotFound`. In hindsight, I wish `ActiveRecord::RecordNotFound` was a default *only* when reported from a Rails server.

I think that with this caveat in mind we could definitely add a global ignore list that just doesn't have any defaults, which eliminates the contextual issue (by leaving the choice of what to globally ignore to the user). What do you guys think?

aaronrenner:
I agree in certain situations it would be nice to ignore specific errors, while in other situations you'd want to report them. In this case, I wanted to ignore `MyAppWeb.UnauthenticatedError`s that were raised specifically within this phoenix endpoint. Also, since umbrella apps share config, a global honeybadger ignore errors configuration might be a pain because you'd end up mixing errors across application boundaries.

While I think there's a bigger discussion to be had about how to ignore errors, the purpose of this PR is to allow overriding `handle_errors/2` so calling `use Honeybadger.Plug` prevent the developer from making additional modifications to how errors are handled. My personal vote is to make this function overridable  and continue the discussion on how to exclude errors from reporting in a separate issue. 

Thanks for all of the thought you've already put into this, and I appreciate the great product you've built!

minhajuddin:
@joshuap  I agree that having a different set of errors to ignore in the plug context makes sense. Adding a plug specific list is also trivial with our current code base. My vote is to currently add ignored errors in the plug context and then add global errors later.

joshuap:
I've decided to merge this PR after taking a look at the [official docs for `Plug.ErrorHandler`](https://hexdocs.pm/plug/Plug.ErrorHandler.html), which provides the `handle_errors` function. Their example defines a public `handle_errors` function.

I am always hesitant to support additional APIs, which is why I wasn't quick to merge this change. I don't mind as much now, since the API actually belongs to `Plug.ErrorHandler` (we just happen to also be using it).

In the future I'd love to add additional options for both globally ignoring errors as well as a Plug-specific config option.

joshuap:
@minhajuddin could you add a CHANGELOG entry and get this merged?

aaronrenner:
@joshuap Thank you for merging this! This will be really helpful.

PR #126, opened 2018-01-04
Add Honeybadger.clear_context

aaronrenner:
I am trying to use `Honeybadger.context/1` inside a [Honeydew](https://github.com/koudelka/honeydew) worker, which is a long-lived process that handles multiple jobs. Because these workers are long-lived, when a worker starts a new job the context from the previous job is carried over until the new job overwrites it.

This adds a `Honeybadger.clear_context/0` function which could be called at any time to completely reset the context for the current process. I can't think of any additional issues this would cause.

I also added docs and typespecs for `context/0` and `context/1`.

### DIFF
--- a/lib/honeybadger.ex
+++ b/lib/honeybadger.ex
@@ -137,6 +137,7 @@ defmodule Honeybadger do
   end
 
   @context :honeybadger_context
+  @type context :: map
 
   @doc false
   def start(_type, _opts) do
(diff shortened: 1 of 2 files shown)


lib/honeybadger.ex
```diff
@@ -209,15 +210,34 @@ defmodule Honeybadger do
     |> Client.send_notice()
   end
 
+  @doc """
+  Retrieves the context that will get sent to the Honeybadger API when/if an
+  exception occurs in the current process.
+  """
+  @spec context :: context
   def context do
     (Process.get(@context) || %{}) |> Enum.into(Map.new)
```
sorentwo: I think we can use `Process.get/2` here, to set a default to `%{}` without using `||`. The value can't ever be set as `nil` because `context/1` would fail.

Additionally, it would be nice to remove the useless `Enum.into/2`.

lib/honeybadger.ex
```diff
@@ -209,15 +210,34 @@ defmodule Honeybadger do
...
   end
 
-  def context(keyword_or_map) do
-    Process.put(@context, Map.merge(context(), Enum.into(keyword_or_map, %{})))
+  @doc """
+  Merges `additional_context` into the the context that will get sent to the
+  Honeybadger API when/if an exception occurs in the current process.
+  """
+  @spec context(map | keyword) :: context
+  def context(additional_context) do
```
sorentwo: I like the improved variable name here. What do you think about adding a guard, as this will blow up without a map or a keyword list? Also, it cements that the way to clear the context would be with `clear_context/0` rather than `context(nil)`.

aaronrenner:
@sorentwo Good ideas. I just pushed a commit that incorporates your feedback.

PR #129, opened 2018-02-15
Use lazy logging in the client module

sorentwo:
This allows the log message to be ignored when the log-level is beyond the custom threshold. For example, successful requests log the API success with a call to `Logger.debug`, which probably won't be enabled in production.

There is a little bit of other cleanup in here as well. I've annotated `GenServer` callbacks and added some type specs, but most of the changes are from running `mix format` on the module.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ CHANGELOG](http://keepachangelog.com/) for how to update this file. This project
 adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
+### Changed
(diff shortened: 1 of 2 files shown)


minhajuddin:
@sorentwo The build seems to be failing for these changes, it seems to be because of the change in the `not` expression. Can you take a look at it.

sorentwo:
@minhajuddin Ah, it's because I used the 1.6 formatter. I've restored it to the `not thing in thing` syntax.

minhajuddin:
Looks great, can you add CHANGELOG entry

sorentwo:
@minhajuddin An entry has been added.

PR #131, opened 2018-03-04
Fix a regression which was causing 404s to be reported

### LINKED ISSUE #130: 404s are sent as errors to Honeybadger
I am using v0.8.0. For some reason we're getting 404s reported as errors.

```
❯ curl -I [redacted-url-that-produces-404]
HTTP/1.1 404 Not Found
[…]
```

The error is coming from Ecto query not finding a thing, but perhaps Honeybadger is failing to match [something here](https://github.com/honeybadger-io/honeybadger-elixir/blob/ee3417585b83d93266b4986084d048a76f13d007/lib/honeybadger/plug.ex#L58) (didn't research this, just guessing from a glance at the code).

We aren't doing anything weird, just have `use Honeybadger.Plug` in the router.

Phoenix 1.3.0, Elixir 1.6.0, Erlang 20

minhajuddin:
Fixes #130

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 - Use lazy logging within `Honeybadger.Client`, this allows compile time purging
   when the log level is set higher.
+### Fixed
+- Fix a regression which was causing `Not Found (404)` errors to be reported.
 
 ## [v0.8.0] - 2018-01-17
 ### Added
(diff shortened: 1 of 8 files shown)


minhajuddin:
@sorentwo Can you do a quick review of this.

minhajuddin:
The `Ecto.NoResultsError` exception doesn't have a `plug_status` attribute. However `:phoenix_ecto` implements `Plug.Exception` which returns a `404` https://github.com/phoenixframework/phoenix_ecto/blob/3e29c636a40ed0e335afddf34e4feb3596498d19/lib/phoenix_ecto/plug.ex#L4. I've added this check using `Plug.Exception.status`, so any exception which implements `Plug.Exception` and returns a `404` will now be ignored.

PR #134, opened 2018-03-17
Cleanup and refactor

minhajuddin:
This has the suggestion from #133 and formatted code. @joshuap we also need to do a release on hex.


lib/honeybadger/client.ex
```diff
@@ -81,7 +81,7 @@ defmodule Honeybadger.Client do
     env_name = get_env(opts, :environment_name)
     excluded = get_env(opts, :exclude_envs)
 
-    not maybe_to_atom(env_name) in excluded
+    not (maybe_to_atom(env_name) in excluded)
```
sorentwo: Nice workaround of this formatter incompatibility 👍

PR #143, opened 2018-05-04
Replace Poison with Jason for JSON encoding

### LINKED ISSUE #136: Make the json decoder configurable
[jason](https://github.com/michalmuskala/jason) is a new json encoder that phoenix supports as well (https://github.com/phoenixframework/phoenix/issues/2693).

It would be nice if this could be configured in honeybadger to use an other json decoder.

### LINKED ISSUE #137: Don't crash if Poison.encode fails
I think it would be useful to not crash when `Poison.encode` fails [here](https://github.com/honeybadger-io/honeybadger-elixir/blob/8c52e0f1a7ad8c4211cd87aed70da6f3e0806e07/lib/honeybadger/client.ex#L114). 

Instead there maybe could be a fallback (`inspect()`) and this could be sent to honeybadger along with an `encoding` error?

If you are open to this, I'll see if I can create a PR.

sorentwo:
This swaps out Poison for Jason, a much faster JSON encoder that will be standard for upcoming versions of Ecto, Phoenix, etc.

Unlike Poison, Jason does _not_ support encoding arbitrary structs—they must derive from `Jason.Encoder`. This adds a derive instruction for the `Notice` struct, which is the primary place where we're actually encoding JSON.

This also switches from using `encode!` to `encode`, allowing us to catch encoding failures and log them as warnings rather than swallow them wholesale.

Fixes #137 
Fixes #136


tcoopman:
Be advised that for the moment Jason can still throw on `encode` https://github.com/michalmuskala/jason/pull/40. Hopefully the pr is accepted soon.

PR #145, opened 2018-05-07
Reduce the log level used for dev mode warning

sorentwo:
With the dev mode warning logging at the `warn` level it is displayed during every test run. This is noisy and unnecessary—it is fully expected that errors aren't reported during dev/test mode.


minhajuddin:
@sorentwo There is a failing test because we assert that a specific warning is emitted, can you please change the test to make it pass :)

sorentwo:
@minhajuddin Test fixed and rebased off of master.

PR #150, opened 2018-05-26
Notify error reports with a malformed stacktrace

### LINKED ISSUE #149: No matching function clause Honeybadger.Backtrace.from_stacktrace/1
Hello,

I'm getting the following application logs:

> {"severity":"warn","msg":"Unable to notify Honeybadger! FunctionClauseError: no function clause matching in Honeybadger.Backtrace.from_stacktrace/1","appname":"phone-filtering"}

From looking at the source code, `stacktrace` [must not be a list](https://github.com/honeybadger-io/honeybadger-elixir/blob/master/lib/honeybadger/backtrace.ex#L22)?

I don't have any other debugging info to go on unfortunately. Our HB setup is the straightforward plug integration:

```
defmodule PhoneFilteringWeb.Router do
  use PhoneFilteringWeb, :router
  use Honeybadger.Plug

```

sorentwo:
The stacktrace in an error report _should_ always be a list, but sometimes a report comes through with another type. In that case we can discard the stacktrace and notify of the exception without it.

Fixes #149


sorentwo:
@minhajuddin Done. Just a heads up, you'll have a conflict with the current CHANGELOG PR.

minhajuddin:
Can you push the commit with the changes to CHANGELOG.md, I don't see them in the diff :slightly_smiling_face:

sorentwo:
@minhajuddin Oops. Yes, amended and pushed.

PR #153, opened 2018-06-18
Fix `excluded_envs` in code snippet in README.md

jasonroelofs:


jasonroelofs:
I ... just realized that I didn't double check which one is actually correct. The README is now consistent but unfortunately totally wrong here. I'll open another to fix this properly.

minhajuddin:
Thanks

PR #155, opened 2018-06-20
Update to Phoenix 1.3.3 and Plug 1.6

chriscarter90:
Upgrades Phoenix to 1.3.3, and Plug to 1.6.

Fixes the issue where the connection's peer is no longer available directly as a field.


chriscarter90:
Hmm, Travis seemed to have a problem building on one of the Elixir/OTP versions. Can it be re-run?

chriscarter90:
Is there any chance we could get this merged?

minhajuddin:
@chriscarter90 I am a bit hesitant on merging this as this is a breaking change (merging this will break code for projects which have an older plug dependency). I see that @josevalim has said that this change will be fixed by adding back the `peer` attribute to the struct here: https://github.com/elixir-plug/plug/issues/722 . Let us hold off on this merge till we get a fix from @josevalim :smile:

chriscarter90:
Ah, I hadn't seen that. In which case, fair enough! 👌

josevalim:
peer will be deprecated anyway. So my suggestion is to check if `get_peer_data` is defined and act accordingly. i.e. you need to do both the behaviour in this PR and the old one.

minhajuddin:
@chriscarter90 Thanks for the PR :smile:, We have a fix for this on master (#157 fixed this)

PR #161, opened 2018-07-02
Add a custom JSON encoder to handle structs in context

### LINKED ISSUE #160: (Protocol.UndefinedError) protocol Jason.Encoder not implemented
A customer reported having the following error after upgrading honeybadger-elixir from `0.9` to `0.10.1`:

> ** (Protocol.UndefinedError) protocol Jason.Encoder not implemented for %App.Repo.Xyz{id: "xyz", ...}, Jason.Encoder protocol must always be explicitly implemented.

minhajuddin:
Fixes #160

### DIFF
--- a/lib/honeybadger/json.ex
+++ b/lib/honeybadger/json.ex
@@ -1,0 +1,53 @@
+defmodule Honeybadger.JSON do
+  def encode(notice) do
+    # try to encode without going the rabbit hole
+    case safe_encode(notice) do
+      {:ok, output} ->
+        {:ok, output}
+
+      {:error, %Protocol.UndefinedError{}} ->
+        notice
+        |> to_encodeable
+        |> Jason.encode()
+    end
+  end
+
+  # struct
+  defp to_encodeable(%_{} = input) do
+    input
+    |> Map.from_struct()
+    |> to_encodeable()
+  end
+
+  # map
+  defp to_encodeable(%{} = input) do
+    for {k, v} <- input, into: %{} do
+      {k, to_encodeable(v)}
+    end
+  end
+
+  # list
+  defp to_encodeable(input) when is_list(input) do
+    for v <- input, do: to_encodeable(v)
+  end
+
+  # tuple
+  defp to_encodeable(input) when is_tuple(input) do
+    input
+    |> Tuple.to_list()
+    |> to_encodeable
+  end
+
+  defp to_encodeable(input) do
+    input
+  end
+
+  def safe_encode(input) do
+    try do
+      Jason.encode(input)
+    rescue
+      e in Protocol.UndefinedError ->
+        {:error, e}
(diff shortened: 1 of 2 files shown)


minhajuddin:
@sorentwo would love a quick review :smile:

lib/honeybadger/json.ex
```diff
@@ -0,0 +1,53 @@
+defmodule Honeybadger.JSON do
```
sorentwo: Can we add a `@moduledoc false` to make it clear that we aren't documenting this one on purpose?

lib/honeybadger/json.ex
```diff
@@ -0,0 +1,53 @@
...
+  def encode(notice) do
```
sorentwo: We don't make any use of dialyzer currently, but I've been trying to include type specs when we introduce or update functions. It would be nice to have one here as well.

lib/honeybadger/json.ex
```diff
@@ -0,0 +1,53 @@
...
+  # struct
+  defp to_encodeable(%_{} = input) do
+    input
+    |> Map.from_struct()
+    |> to_encodeable()
+  end
+
+  # map
+  defp to_encodeable(%{} = input) do
+    for {k, v} <- input, into: %{} do
```
sorentwo: We don't use single letter variables in other places, it would be great to avoid them here (and elsewhere in this module) as well.

lib/honeybadger/json.ex
```diff
@@ -0,0 +1,53 @@
...
+  end
+
+  # tuple
+  defp to_encodeable(input) when is_tuple(input) do
+    input
+    |> Tuple.to_list()
+    |> to_encodeable
+  end
+
+  defp to_encodeable(input) do
```
sorentwo: I think we still need a clause for `pid`, `reference` and `port` here as well. There isn't any reason they _should_ be stashed inside of context, but for the sake of completeness I think we should handle them. Something just to inspect them, like:

```elixir
defp to_encodeable(input) when is_pid(input) or is_port(input) or is_reference(input) do
  inspect(input)
end
```

test/honeybadger/json_test.exs
```diff
@@ -0,0 +1,48 @@
...
+
+  alias Honeybadger.{Notice, JSON}
+
+  describe "encode" do
+    test "encodes notice" do
+      notice = Notice.new(%RuntimeError{message: "oops"}, %{}, [])
+      assert JSON.encode(notice) == Jason.encode(notice)
+    end
+
+    defmodule Req do
```
sorentwo: Can we move this outside of the `describe` call? It always trips me up when functions or modules are declared within a `describe` because it looks like they are scoped to the block but they are actually scoped to the module.

PR #168, opened 2018-07-25
Remove use of System.stracktrace/0 outside of try

sorentwo:
This adds tests for Erlang 21.0 and Elixir 1.7.0 to the build matrix. One of the big changes is the introduction of `__STACKTRACE__` and the deprecation of `System.stacktrace/0`.

To prevent compile time warnings moving forward we are removing the fallback `System.stacktrace/0` call inside of `honeybadger.ex`. The call was made outside of a `try/catch` and wasn't guaranteed to return the correct stacktrace anyhow.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,5 +23,6 @@ matrix:
       elixir: 1.5
     - otp_release: 20.1
       elixir: 1.6
-            
+    - otp_release: 21.0
+      elixir: 1.7
 sudo: false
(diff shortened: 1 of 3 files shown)


minhajuddin:
@sorentwo can you check the failing test, it is failing across all elixir versions :slightly_smiling_face:

sorentwo:
@minhajuddin Switched the `backtrace/1` clauses around to catch the malformed stack trace. That did the trick.

I really wish Travis would notify me about broken builds!

PR #174, opened 2018-08-09
Guard against nil returns from Process.info/2

### LINKED ISSUE #172: ErlangError: :badmatch
Here's a screenshot of the Honeybadger stacktrace:

2018-08-06-121619_1151x912_scrot

Running honeybadger-elixir version 0.10.2 on Elixir 1.6.

sorentwo:
When a process is no longer alive any calls to `Process.info/2` will return `nil` instead of the expected tuple. Inside of the fallback clause for `backtrace/1` we always matched on a stacktrace tuple and didn't account for the possibility of a `nil` return. This change makes the backtrace generation safer by using a `case` instead of a fixed match.

During my experimentation I wasn't able to call `backtrace/1` from a process and have the process die before `Process.info/2` was called. The case clause is defensive and should cover the known failure case, but it seems to be untestable.

Closes #172

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ CHANGELOG](http://keepachangelog.com/) for how to update this file. This project
 adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
+### Fixed
+- Prevent crashes reporting `:badmatch` when fetching the current stacktrace
+  from a dead process.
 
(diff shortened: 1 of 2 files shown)


sorentwo:
@thmzlt Would you be able to try running off this branch? As I noted in the commit message, I'm unable to replicate this with tests or in the console. During development `badmatch` is converted into a more helpful Elixir `MatchError` anyhow, making it even trickier.

Issue #195, opened 2019-01-16
`PlugData` does not handle invalid remote IP 


van-mronov:
We noticed in our system some interesting bug.
One of our internal plug put remote IP in `Plug.Conn` structure as string not as tuple.
When error occurs in controller honeybadger tried to parse this ip string with `:inet.ntoa/1`. It expects tuple not string and returns `{:error, :einval}` which honeybadger tried to pass into `List.to_string/1` and it crashed.

`lib/honeybadger/plug_data.ex`:
```elixir
defp remote_addr(%Conn{remote_ip: remote_ip}) do
  remote_ip
  |> :inet.ntoa()
  |> List.to_string()
end
```

I know that the root cause is not in honeybadger. But maybe it makes sense to handle `{:error, :einval}` and try to convert remote ip address to string, for example, or maybe just return ip address as `"unknown"`:
```elixir
defp remote_addr(%Conn{remote_ip: remote_ip}) do
  case :inet.ntoa(remote_ip) do
    {:error, :einval} ->
      to_string(remote_ip)

    charlist ->
      List.to_string(charlist)
  end
end
```

van-mronov:
Fixed

PR #196, opened 2019-01-16
Try to convert remote IP to string in case of parsing failure

van-mronov:
Fix for #195


joshuap:
Looks good, thanks @van-mronov!

PR #199, opened 2019-01-26
Change `Map.t()` to `map`

### LINKED ISSUE #198: Map.t() isn't a valid dialyzer type

mjs2600:
This caused issues with Dialyzer since `Map.t()` isn't a type defined
in the language. See https://hexdocs.pm/elixir/typespecs.html

Closes #198


sorentwo:
Thanks for the fix. We'll get dialyzer hooked up for CI runs soon.

Issue #200, opened 2019-02-07
Erlang error: :case_clause from Honeybadger


axelson:
Getting this on 0.10.2 and 0.10.3:

Stacktrace:
```
process.ex  662 info/2
lib/process.ex:662:in `info/2'
honeybadger.ex  316 backtrace/1
lib/honeybadger.ex:316:in `backtrace/1'
honeybadger.ex  209 notify/3
lib/honeybadger.ex:209:in `notify/3'
logger.ex  54 handle_error/1
lib/honeybadger/logger.ex:54:in `handle_error/1'
logger.ex  21 handle_event/2
lib/honeybadger/logger.ex:21:in `handle_event/2'
gen_event.erl  577 server_update/4
gen_event.erl:577:in `server_update/4'
gen_event.erl  559 server_notify/4
gen_event.erl:559:in `server_notify/4'
```

Honeybadger link for any employees: https://app.honeybadger.io/projects/50379/faults/39800245#notice-summary

joshuap:
Thanks for the report, @axelson. Do you have any code that can reproduce the error, by any chance?

axelson:
No I don't sorry. Nothing in the stacktrace is our code so I'm not really sure what's happening.

bernardd:
We're seeing the same thing, but we're generally able to extract the correct call stack from our logging system. In our most recent example the code looked like this:

```elixir
  defp send_invitation(
         %{e164_phone_number: number} = r,
         sent_numbers,
         requestor
       )
       when not is_nil(number) do
    r =
      with {:ok, body} <- sms_invitation_body(requestor),
           :ok <- Messenger.send(number, body, requestor) do
        Map.put(r, :result, :external_invitation_sent)
      else
        {:error, e} ->
          r
          |> Map.put(:result, :sms_error)
          |> Map.put(:error, inspect(e))
      end
```

The key bit is the `with` clause - the `Messenger.send` function returned `{:error, "Some string or other", 404}` which matched neither the success case nor the single pattern under `else`, triggering a `with_clause` exception.

We're running erlang 21 (which I just noticed it a couple of minor releases behind - I should fix that) and Elixir 1.7.1. Presuming you can see our account, the error is https://app.honeybadger.io/projects/50679/faults/40249305. Please let me know if I can provide any more info.

axelson:
@bernardd thanks for that, your report is much more helpful than mine :)

[...]

PR #204, opened 2019-02-18
Migrate to Elixir 1.7+ Logger

sorentwo:
This PR is a multi-pronged attempt to provide richer context, more consistent reporting and eliminate inconsistent stacktraces that have plagued so many of us (see #200 for an example).

* [SASL Logging](http://erlang.org/doc/apps/sasl/error_logging.html) — This switches from using the deprecated `:error_logger` to the enhanced `Logger` backend provided since Elixir 1.7+ 
* [Logger Backend](https://hexdocs.pm/logger/Logger.html#module-metadata) — Context now uses the Logger metadata for storage. This is because the new logger messages/meta doesn't include a `:dictionary` key to extract the metadata from, and the process isn't guaranteed to be alive so `:process_info` isn't reliable either. As a bonus, we get very consistent metadata formatting.
* [\_\_STACKTRACE\_\_](https://hexdocs.pm/elixir/Kernel.SpecialForms.html#__STACKTRACE__/0) — Calling `notify/3` without a stacktrace (or with `[]`) no longer pulls `:current_stacktrace` from the current process. This directly resulted in incorrect backtraces that made error reports hard to work with. Calling `System.stacktrace/0` is deprecated and [get_stacktrace](http://erlang.org/doc/man/erlang.html#get_stacktrace-0) will be removed from future OTP versions.
* Enhanced context for `GenServer`, `Task`, `gen_event`, and `spawn` reporting. By parsing the message format we now get `last_message` and `state` for `Gen*` reports, and `function/args` for `Task` reports. The `Logger` tests have been modified to cover these additional cases.

There aren't any breaking API changes, but the package now requires Elixir 1.7+ and OTP 21+. The project's elixir version and build matrix have been modified accordingly, but we'll need some additional messaging and an upgrade guide for people upgrading.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,22 +1,10 @@
 language: elixir
 sudo: false
 elixir:
-  - 1.4
-  - 1.5
-  - 1.6
   - 1.7
   - 1.8
 otp_release:
-  - 20.3
   - 21.2
-matrix:
-  exclude:
-    - otp_release: 21.2
-      elixir: 1.4
-    - otp_release: 21.2
-      elixir: 1.5
-    - otp_release: 21.2
-      elixir: 1.6
 env:
   - MIX_ENV=test
 script:
--- a/lib/honeybadger.ex
+++ b/lib/honeybadger.ex
@@ -36,8 +36,9 @@ defmodule Honeybadger do
         raise RunTimeError, message: "Oops"
       rescue
         exception ->
-          metadata = %{user_id: 1, account: "A Very Important Customer"}
-          Honeybadger.notify(exception, metadata)
+          context = %{user_id: 1, account: "A Very Important Customer"}
+
+          Honeybadger.notify(exception, context, __STACKTRACE__)
       end
 
   Note that `notify` may be used outside of `try`, but it will use a different
@@ -163,13 +164,19 @@ defmodule Honeybadger do
   This is the primary way to do manual error reporting and it is also used
   internally to deliver logged errors.
 
+  ## Stacktrace
+
+  Accessing the stacktrace outside of a rescue/catch is deprecated. Notifiations should happen
+  inside of a rescue/catch block so that the stacktrace can be provided with `__STACKTRACE__`.
+  Stacktraces _must_ be provided and won't be automatically extracted from the current process.
+
   ## Example
 
       try do
         do_something_risky()
       rescue
         exception ->
-          Honeybadger.notify(exception)
+          Honeybadger.notify(exception, %{}, __STACKTRACE__)
       end
 
   Send a notification directly from a string, which will be sent as a
@@ -185,7 +192,7 @@ defmodule Honeybadger do
 
   Send a notification as a `badarg` atom:
 
-      iex> Honeybadger.notify(:badarg, %{})
+      iex> Honeybadger.notify(:badarg)
       :ok
 
   If desired additional metadata can be provided as well:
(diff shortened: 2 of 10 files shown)


sorentwo:
The failures are coming from a [dynamic logger issue](https://github.com/elixir-lang/elixir/pull/8424) that is present in 1.7, but was fixed in 1.8. Working on it.

sorentwo:
@joshuap Everything is fixed up now. We're going to run this branch in staging at dscout this week to test things and check out the context reporting.

lib/honeybadger.ex
```diff
@@ -160,26 +158,25 @@ defmodule Honeybadger do
...
-
   @doc """
   Send an exception notification, if reporting is enabled.
 
   This is the primary way to do manual error reporting and it is also used
   internally to deliver logged errors.
 
+  ## Stacktrace
+
+  Accessing the stacktrace outside of a rescue/catch is deprecated. Notifiations should happen
```
starrhorne: Do people currently use `Honeybadger.notify` outside of rescue/catch blocks? If so, do we need to let them know that they should stop?

.travis.yml
```diff
@@ -1,22 +1,10 @@
 language: elixir
 sudo: false
 elixir:
-  - 1.4
```
starrhorne: What happens if people are using the now-unsupported versions of Elixir? Do they get an error when running `mix deps.get`? Or does the code just break?

lib/honeybadger.ex
```diff
@@ -195,48 +192,62 @@ defmodule Honeybadger do
...
   """
-  @spec context(map | keyword) :: context
-  def context(additional_context)
-      when is_map(additional_context) or is_list(additional_context) do
-    Process.put(@context, Map.merge(context(), Enum.into(additional_context, %{})))
+  @spec context(map() | keyword()) :: map()
+  def context(map) when is_map(map), do: context(Keyword.new(map))
+
+  def context(keyword) do
+    Logger.metadata(keyword)
```
joshuap: Does setting context work like before if I'm notifying manually from the current process? I.e., if I synchronously call `Honeybadger.context(%{foo:"bar"})`, and then `Honeybadger.notify("testing")`, the context will be reported with the error (and I assume no stack trace)?

lib/honeybadger.ex
sorentwo: People may. Using `Honeybadger.notify/3` will still work, there just won't be a stacktrace included. For what it's worth, Sentry and Rollbax work the same way.

.travis.yml
sorentwo: If my understanding of hex dependency resolution is correct, because the package declares the minimum version they'll be stuck at a `0.10.X` version and unable to move to `>= 0.11`.

lib/honeybadger.ex
sorentwo: It works just like it did before. We were using the process dictionary to store context, and that is exactly how the logger stores metadata.

sorentwo:
@joshuap Thanks for taking a look. Here are the important changes that users would need to know when upgrading:

1. Upgrading to `0.11` and above will require Elixir 1.7+ and Erlang/OTP 21+
2. Manually calling `Honeybadger.notify/3` will not include a stacktrace by default
3. A stacktrace can be provided from within a `rescue/catch` block via the `__STACKTRACE__` macro, e.g. `Honeybadger.notify(SomeError, %{my: :context}, __STACKTRACE__)`

joshuap:
Thanks @sorentwo! I'll let you guys run the branch for a bit--lmk when you feel it's good to go and I'll merge, release, and write up a blog post with that info (you are also welcome to guest post on our blog if you would like).

sorentwo:
@joshuap Alright, we're satisfied with the changes and ready to start running this branch in production. Merge whenever you are ready and we'll run on master next week (we normally do anyhow).

joshuap:
v0.11.0 has been released

sorentwo:
@joshuap Fantastic! We've been running in production since Monday and everything has gone smoothly.

Issue #207, opened 2019-03-01
Pass revision of current deploy?


thomasbrus:
Is it possible to pass the revision of the current deploy? I couldn't find anything in the docs or in the code. 

Both the Ruby & JS version support a `:revision` config option: 
- https://docs.honeybadger.io/ruby/gem-reference/configuration.html
- https://docs.honeybadger.io/lib/javascript/reference/configuration.html

joshuap:
@thomasbrus it doesn't look like we support this feature yet. I'm open to PRs for it, though. I think it would require adding a `revision` configuration option, and then sending that option in the API payload similar to [how we do for Ruby](https://github.com/honeybadger-io/honeybadger-ruby/blob/35d210bb2a97ab3d19c892defa40c0854a0ebe50/lib/honeybadger/notice.rb#L215).

thomasbrus:
I'll have a look at it soon-ish!

PR #208, opened 2019-03-04
Added :revision configuration option

thomasbrus:
Closes honeybadger-io/honeybadger-elixir#207

- Could still fail if project isn't a git repository, should we try/catch this?
- My editor (prettier plugin) added some newlines here and there, hope that's alright

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ CHANGELOG](http://keepachangelog.com/) for how to update this file. This project
 adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
+### Added
+- Added `:revision` configuration option
 
 ## [v0.11.0] - 2019-02-28
 When upgrading to v0.11, users should be aware of a few important changes:
--- a/lib/honeybadger.ex
+++ b/lib/honeybadger.ex
@@ -21,6 +21,7 @@ defmodule Honeybadger do
         proxy: "http://proxy.net:PORT",
         proxy_auth: {"Username", "Password"},
         project_root: "/home/skynet",
+        revision: System.get_env("GIT_REVISION"),
         use_logger: true,
         notice_filter: Honeybadger.NoticeFilter.Default,
         filter: Honeybadger.Filter.Default,
@@ -302,6 +303,7 @@ defmodule Honeybadger do
     config
     |> Keyword.put_new_lazy(:hostname, hostname)
     |> Keyword.put_new_lazy(:project_root, &File.cwd!/0)
+    |> Keyword.put_new(:revision, nil)
   end
 
   defp verify_environment_name!(config) do
--- a/README.md
+++ b/README.md
@@ -161,7 +161,8 @@ If you'd rather read, eg., `environment_name` from the OS environment, you can d
 
 ```elixir
 config :honeybadger,
-  environment_name: {:system, "HONEYBADGER_ENV"}
+  environment_name: {:system, "HONEYBADGER_ENV"},
+  revision: {:system, "HEROKU_SLUG_COMMIT"}
 ```
 
 _NOTE: This works only for the string options, and `environment_name`._
@@ -177,6 +178,7 @@ Here are all of the options you can pass in the keyword list:
 | `hostname`               | Hostname of the system your application is running on                                         | `:inet.gethostname`                      |
 | `origin`                 | URL for the Honeybadger API                                                                   | `"https://api.honeybadger.io"`           |
 | `project_root`           | Directory root for where your application is running                                          | `System.cwd/0`                           |
+| `revision`               | The project's git revision                                                                    | `nil`                                    |
 | `filter`                 | Module implementing `Honeybadger.Filter` to filter data before sending to Honeybadger.io      | `Honeybadger.Filter.Default`             |
 | `filter_keys`            | A list of keywords (atoms) to filter.  Only valid if `filter` is `Honeybadger.Filter.Default` | `[:password, :credit_card]`              |
 | `filter_args`            | If true, will remove function arguments in backtraces                                         | `true`                                   |
--- a/lib/honeybadger/notice.ex
+++ b/lib/honeybadger/notice.ex
@@ -13,7 +13,12 @@ defmodule Honeybadger.Notice do
 
   @typep notifier :: %{name: String.t(), url: String.t(), version: String.t()}
 
-  @typep server :: %{environment_name: atom, hostname: String.t(), project_root: Path.t()}
+  @typep server :: %{
+           environment_name: atom,
+           hostname: String.t(),
+           project_root: Path.t(),
+           revision: String.t()
+         }
 
   @type t :: %__MODULE__{
           notifier: notifier,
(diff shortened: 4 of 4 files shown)


CHANGELOG.md
```diff
@@ -24,81 +30,103 @@ When upgrading to v0.11, users should be aware of a few important changes:
...
 ### Changed
+
 - Replace Poison with Jason for JSON encoding.
 - Reduce the log level used for dev mode warning.
 - Update dependenices.
 
 ### Fixed
 
-  JSON encoding error.
+JSON encoding error.
```
sorentwo:
```suggestion
- JSON encoding error.
```

CHANGELOG.md
```diff
@@ -6,14 +6,20 @@ adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Added
+
```
sorentwo: I appreciate the extra whitespace in here 👍

lib/honeybadger.ex
```diff
@@ -14,6 +14,7 @@ defmodule Honeybadger do
       config :honeybadger,
         api_key: "mysupersecretkey",
         environment_name: :prod,
+        revision: "710a4d9bb5afdf31c156d042c7f48f3c107d30e1",
```
sorentwo: This isn't something we'd expect people to add manually, so I don't think be listed as an example.

lib/honeybadger.ex
```diff
@@ -299,9 +300,17 @@ defmodule Honeybadger do
       |> List.to_string()
     end
 
+    git_revision = fn ->
```
sorentwo: If `git` isn't available on the system this will raise `** (ErlangError) Erlang error: :enoent`. We should use [find_executable/1](https://hexdocs.pm/elixir/System.html#find_executable/1) to check for the executable before making the call.

thomasbrus: @sorentwo Oh I see yes, I just noticed that it would throw an actual error then yes

thomasbrus:
@sorentwo Alright I pushed some changes. I don't have much time to implement the heroku slug thing although I agree it'd be nice, perhaps we could leave that for another time? Furthermore I used with + `which git` to check if git is available, seemed a bit more elegant. Let me know if you have any other comments

lib/honeybadger.ex
```diff
@@ -299,9 +299,19 @@ defmodule Honeybadger do
...
+      with {_path, 0} <- System.cmd("which", ["git"]),
```
sorentwo: Using `with` is great, but we can't rely on `which` for a windows system. While it is doubtful that anybody would be deploying on windows, they may be developing on it.

```suggestion
      with <<_path>> <- System.find_executable("git"),
```

Additionally, what do you think about moving this into a private function? There is a lot happening in the code block.

lib/honeybadger.ex
```diff
@@ -299,9 +299,19 @@ defmodule Honeybadger do
...
+           {revision, 0} <- System.cmd("git", ["rev-parse", "HEAD"]) do
+        String.trim_trailing(revision)
+      else
+        {_stdout, _code} -> nil
```
sorentwo:
```suggestion
        _ -> nil
```

lib/honeybadger.ex
thomasbrus: @sorentwo that doesn't work, how about: `with path when is_binary(path) <- System.find_executable("git")` ? cause it returns nil otherwise.

sorentwo: Quite right, my mistake! That looks great. 👍

thomasbrus:
Alright I think I got it all now @sorentwo

CHANGELOG.md
joshuap: I hate to be the one to nitpick, but the old whitespace actually conforms to the [Keep a Changelog spec](https://keepachangelog.com/en/1.0.0/), which is why there wasn't any. We use the same format across all of our projects, so changing it for just this repos doesn't feel right to me. Also, it doesn't feel right to change it as a result of an unrelated PR. Can we remove the whitespace? Thanks!

CHANGELOG.md
```diff
@@ -6,14 +6,20 @@ adheres to [Semantic Versioning](http://semver.org/).
...
+- Added `:revision` configuration option which is automatically set to the current git revision by default
+
 ## [v0.11.0] - 2019-02-28
+
 When upgrading to v0.11, users should be aware of a few important changes:
 
 - You must be on Elixir 1.7+ and Erlang/OTP 21+
 - Due to the deprecation of `System.stacktrace/0` and the introduction of [`__STACKTRACE__`](https://hexdocs.pm/elixir/Kernel.SpecialForms.html#__STACKTRACE__/0), manually calling `Honeybadger.notify/3` will no longer include a stacktrace by default. See the [issue discussion](https://github.com/honeybadger-io/honeybadger-elixir/pull/204) for more details
-- A stacktrace can be manually provided from within a rescue/catch block via the __STACKTRACE__ macro, e.g. `Honeybadger.notify(SomeError, %{my: :context}, __STACKTRACE__)`
+- A stacktrace can be manually provided from within a rescue/catch block via the **STACKTRACE** macro, e.g. `Honeybadger.notify(SomeError, %{my: :context}, __STACKTRACE__)`
```
joshuap: `__STACKTRACE__` is a literal (the underscores are a part of the name). If this wasn't displaying properly, we could add the backticks around it to make it appear as code:

```suggestion
- A stacktrace can be manually provided from within a rescue/catch block via the `__STACKTRACE__` macro, e.g. `Honeybadger.notify(SomeError, %{my: :context}, __STACKTRACE__)`
```

thomasbrus: Of course no problem (regarding the whitespace)

thomasbrus:
@joshuap Hmm good point, in fact I just deployed this branch in my own app and it did indeed not work (not only because I happen to not have git available in the container where the app is run), but also because there's no .git directory after using `mix release`. So I think that when building the release the git revision would need to be passed in as an ENV var (`HONEYBADGER_REVISION`) by the end user, and then we could read from that var similarly to how @sorentwo noted in https://github.com/honeybadger-io/honeybadger-elixir/pull/208#pullrequestreview-210993302.

thomasbrus:
Okay I removed the whitespace changes. And I changed the approach a bit. If the git revision is available during runtime the user can use set :revision to `{:system, "SOME_ENV_VAR"}` (`"HEROKU_SLUG_COMMIT"` for example).

If the git revision is only available during build time (for example when deploying to AWS), then the user can set :revision to `System.get_env("GIT_REVISION")` in config/prod.exs. They would need to make sure that the env var is present during build time.

Would this be okay for now? In any case I don't think the git revision can be reliably determined in the put_dynamic_env/2 function because when building for production the source files (and .git) dir are no longer present.

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ CHANGELOG](http://keepachangelog.com/) for how to update this file. This project
 adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
+### Added
+- Added `:revision` configuration option
```
sorentwo: Some more details about what this is would be helpful.

PR #216, opened 2019-04-04
Upgrade plug to 1.8.0 and phoenix to 1.4.3

sorentwo:
This upgrades plug and phoenix along with a small change in the plug tests. Plug errors are now wrapped to preserve context and we're reporting the full wrapped error rather than the sub-error, called the "reason".

Supersedes #212 and #213


joshuap:
@sorentwo would we potentially want to start reporting the reason in addition to the wrapper? Our API spec does support "causes", which is an array of additional errors. Currently used by Ruby and Java.

https://docs.honeybadger.io/api/exceptions.html (see example in API payload)

sorentwo:
@joshuap That seems like a great idea. Mind if I tackle that in a separate PR?

joshuap:
Of course! I'll even merge this one. :)

PR #219, opened 2019-05-02
Automatically call Exception.blame/3 for enhanced errors

### LINKED ISSUE #211: Is there a reason not to call `Exception.blame` automatically?
Is there a specific reason that `Exception.blame/3` is not called automatically by Honeybadger? It does say that "This operation is potentially expensive, as it reads data from the file system, parses beam files, evaluates code and so on." but I think this should at least be optional when using honeybadger since it provides a much richer error.

It can convert an error like this:
```
no function clause matching in Messages.handle_message_created/2
```

into:
```
no function clause matching in Messages.build_notifications/3

The following arguments were given to Messages.build_notifications/3:

    # 1
    %Chat.Message{__meta__: #Ecto.Schema.Metadata<:loaded, "messages">, ...snip large struct value}

    # 2
    nil

    # 3
    nil

Attempted function clauses (showing 3 out of 3):

    def build_notifications(message, thread, -%Messages.MessageContainer.MessageContainerImpl{container_id: post_id}-)
    def build_notifications(%{id: message_id}, _thread, -%Communities.Post{message_id: message_id}-)
    def build_notifications(message, thread, -%Communities.Post{} = post-)
```

Might be related to #123

axelson:
It can convert an error like this:

no function clause matching in Messages.handle_message_created/2
into:

no function clause matching in Messages.build_notifications/3

The following arguments were given to Messages.build_notifications/3:

    # 1
    %Chat.Message{__meta__: #Ecto.Schema.Metadata<:loaded, "messages">, ...snip large struct value}

    # 2
    nil

    # 3
    nil

Attempted function clauses (showing 3 out of 3):

    def build_notifications(message, thread, -%Messages.MessageContainer.MessageContainerImpl{container_id: post_id}-)
    def build_notifications(%{id: message_id}, _thread, -%Communities.Post{message_id: message_id}-)
    def build_notifications(message, thread, -%Communities.Post{} = post-)

Fixes #211

### DIFF
--- a/lib/honeybadger.ex
+++ b/lib/honeybadger.ex
@@ -203,6 +203,7 @@ defmodule Honeybadger do
   """
   @spec notify(Notice.noticeable(), map(), list()) :: :ok
   def notify(exception, metadata \\ %{}, stacktrace \\ []) do
+    {exception, _} = Exception.blame(:error, exception, stacktrace)
     backtrace = Backtrace.from_stacktrace(stacktrace)
 
     exception
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ adheres to [Semantic Versioning](http://semver.org/).
 ## [Unreleased]
 ### Added
 - Added `:revision` configuration option
+- Automatically call `Exception.blame/3` to record additional information in the exception
 
 ## [v0.11.0] - 2019-02-28
 When upgrading to v0.11, users should be aware of a few important changes:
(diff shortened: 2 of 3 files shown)


lib/honeybadger.ex
```diff
@@ -203,6 +203,7 @@ defmodule Honeybadger do
   """
   @spec notify(Notice.noticeable(), map(), list()) :: :ok
   def notify(exception, metadata \\ %{}, stacktrace \\ []) do
+    {exception, _} = Exception.blame(:error, exception, stacktrace)
```
sorentwo: The first argument to `blame/3` is "kind", which isn't always `:error`. However, at this level we don't know whether it is `:error`, `:throw`, or  `{:EXIT, pid}`, so hard coding `:error` works for me.

axelson:
Awesome, thanks!

PR #226, opened 2019-06-06
Normalize string errors before exception blame

### LINKED ISSUE #225: Exception.blame being called on notify when noticeable is not an Exception
After upgrading to v0.12, When Notify is called with a string, eg `Honeybadger.notify("some message")`,  the message is being reported as an Erlang Error. 

Should `Exception.blame` be called when the exception parameter is not an Exception?
Note: the parameter can be of type `Notify.noticeable{}` which includes Exception, string, map or atoms.

sorentwo:
We need to convert string based errors to an exception before calling `Exception.blame/3` on exceptions. Calling `blame` requires a proper stacktrace to work. However, normalization took place inside the `Notice` module which was only passed the Honeybadger specific backtrace.

In order to centralize normalization and improve the blame behavior all of the conversion now happens in `Notice.new/3`. As part of this effort the typespecs, guards and function signatures have been improved.

The `Backtrace` and `Notice` modules are also no longer public. There wasn't any reason for developers to call them directly and the small documentation that was there wasn't especially helpful.

Closes #225

@joshuap Nothing critical here, but it seems like something we should get out sooner than later.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ CHANGELOG](http://keepachangelog.com/) for how to update this file. This project
 adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
+#### Fixed
+- Call `Exception.blame/3` after normalizing string errors, which prevents reporint all string messages as Erlang errors. (#225)
 
 ## [v0.12.0] - 2019-05-30
 ### Added
--- a/lib/honeybadger.ex
+++ b/lib/honeybadger.ex
@@ -125,7 +125,7 @@ defmodule Honeybadger do
 
   use Application
 
-  alias Honeybadger.{Backtrace, Client, Notice}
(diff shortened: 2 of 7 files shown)


sorentwo:
@joshuap Bump. Can you give this a look over?

joshuap:
Yeah, I'll take a look at it later this afternoon or tomorrow morning. Thanks!

joshuap:
@sorentwo forgot, would you mind adding a changelog entry? https://github.com/honeybadger-io/honeybadger-elixir/blob/master/CHANGELOG.md

I'll merge/release after that. Does 0.12.1 work for you?

sorentwo:
@joshuap CHANGELOG entry added under "Unreleased". Bumping the patch to v0.12.1 sounds right to me.

joshuap:
Thanks!

PR #235, opened 2019-08-07
Breadcrumbs

rabidpraxis:
# Breadcrumbs

The current functionality includes:

* `Honeybadger.add_breadcrumb` function that appends a breadcrumb into a set of breadcrumbs, stored in the process dictionary (Logger metadata, to be exact)
* Breadcrumbs injected into Notice metadata
* Breadcrumb Collector that delegates to a Ring Buffer collection (with a default of 40 items)
* Sanitized breadcrumb metadata to ensure we are only sending a map 1 level deep
* Breadcrumb filtering added to `Honeybadger.Filter` callbacks
* Basic telemetry setup for capturing Phoenix events

I ran into a few unexpected challenges while injecting breadcrumbs into Logger errors. On the first iteration, I was relying on the Process dictionary to store the breadcrumb set. This worked fine until I realized that `Logger` errors were reported in a different process from the process throwing the error. Context gets around this by using the `Logger.metadata` (as that gets passed to the `handle_event` callback), so I refactored to use the Logger metadata as well. This required me to provide a way to add breadcrumbs to a provided set as well, instead of assuming the set is located somewhere globally, which is why there are multiple arity implementations of some functions.

#### Automatic Breadcrumbs

I have decided to focus on only telemetry events for this PR, as that seems to be the future of instrumentation. I have a few more phoenix events to add, and fit some testing around the Telemetry module.

@sorentwo A general review would be great. I have also been banging my head on this one for a while and have decided to seek help, any insight would be appreciated. I want to use `telemetry` to collect automatic breadcrumb Ecto events. I can't for the life of me figure out how to make sure that I have the correct `telemetry_prefix`. If we knew the user's module that includes `Ecto.Repo` then we can get it from the config, but I don't know how to dynamically determine which module has it included.

@joshuap if you want to give it a look as well, that would be 🕶


sorentwo:
Finishing up vacation today and I can review this sometime tomorrow. Thanks for all the details!

lib/honeybadger/filter/mixin.ex
```diff
@@ -36,15 +36,21 @@ defmodule Honeybadger.Filter.Mixin do
...
 
       @doc false
       def filter_map(map) do
         filter_map(map, Honeybadger.get_env(:filter_keys))
       end
 
       def filter_map(map, keys) when is_list(keys) do
-        filter_keys = Enum.map(keys, &canonicalize(&1))
-        drop_keys = Enum.filter(Map.keys(map), &Enum.member?(filter_keys, canonicalize(&1)))
+        filter_keys = Enum.map(keys, &Honeybadger.Utils.canonicalize(&1))
```
sorentwo:
```suggestion
        filter_keys = Enum.map(keys, &Honeybadger.Utils.canonicalize/1)
```

lib/honeybadger/logger.ex
```diff
@@ -65,6 +67,19 @@ defmodule Honeybadger.Logger do
 
   ## Helpers
 
+  defp notify(reason, metadata, stacktrace) do
```
sorentwo: This is doing the same work as `Honeybadger.notify/3` (including overwriting the breadcrumb metadata). Can it be extracted somewhere to make it easier to test and prevent subtle differences?

lib/honeybadger/notice_filter/default.ex
```diff
@@ -28,6 +29,10 @@ defmodule Honeybadger.NoticeFilter.Default do
     Map.put(error, :message, filter.filter_error_message(message))
   end
 
+  defp filter_breadcrumbs(breadcrumbs, filter) do
+    Map.update(breadcrumbs, :trail, %{}, &filter.filter_breadcrumbs(&1))
```
sorentwo:
```suggestion
    Map.update(breadcrumbs, :trail, %{}, &filter.filter_breadcrumbs/1)
```

lib/honeybadger/utils.ex
```diff
@@ -17,4 +17,77 @@ defmodule Honeybadger.Utils do
     |> Module.split()
     |> Enum.join(".")
   end
+
+  @doc """
+  Transform value into a consistently cased string representation
```
sorentwo: This seems like a good place for doctests.

lib/honeybadger/utils.ex
```diff
@@ -17,4 +17,77 @@ defmodule Honeybadger.Utils do
...
+  end
+
+  @doc """
+  Configurable data sanitization. This currently:
+
+  - recursively truncates deep structures (to a depth of 20)
+  - constrains large string values (to 64k)
+  - filters out any map keys that might contain sensitive information.
+  """
+  @default_max_string_size 65526
```
sorentwo: This may be more obvious through multiplication:

```suggestion
  @default_max_string_size 64 * 1024
```

lib/honeybadger/utils.ex
```diff
@@ -17,4 +17,77 @@ defmodule Honeybadger.Utils do
...
+  defp sanitize_val(v, %{depth: depth, max_depth: depth}) when is_map(v) or is_list(v) do
+    @depth_token
+  end
+
+  defp sanitize_val(%{__struct__: _} = struct, opts) do
+    sanitize_val(Map.from_struct(struct), opts)
+  end
+
+  defp sanitize_val(v, %{depth: depth, filter_keys: filter_keys} = opts) when is_map(v) do
+    Enum.reduce(v, %{}, fn {key, value}, acc ->
```
sorentwo: This is a good place to use a `for` comprehension with `into`:

```elixir
for {key, val} <- map, into: %{} do
  if MapSet.member?(filter_keys, canonicalize(key)) do
    {key, @filtered_token}
  else
    {key, sanitize_val(val, Map.put(opts, :depth, depth + 1))}
  end
end

**Comment by rabidpraxis** [comment]
@sorentwo Woah, thank you for the thorough review! I skimmed these and there are some great suggestions. I am off today, so I will dig into these on Monday.

File: lib/honeybadger/logger.ex (hunk 2 above)
Comment (rabidpraxis) [reply lib/honeybadger/logger.ex]: I agree they are doing some similar operations, but they seemed different enough that I wasn't sure unifying them would be worth it. Subtle things like the fact that the lookup keys are different, one returns a different output, they store different breadcrumbs, and one makes sure a key is removed.

I'm open to suggestions if it seems like there is a way to simplify.

**Comment by rabidpraxis** [comment]
@sorentwo Thanks again for the review, I learned a few things!

The ets table sounds interesting. I don't have a ton of experience with ets to know the immediate tradeoffs, but I am assuming we would use it as a k/v store based on pid, with the value being the buffer implementation? That seems like it could also give us options around inspecting breadcrumbs from other processes, which could have some interesting consequences. We would then need to make sure the data is purged after the notice has been generated, right?

Also, I am still unsure of the right way to hook into the Ecto telemetry (making sure we are using correct event name, since they seem to prepend the phoenix app name at the beginning). Do you have any ideas on how to get that telemetry event name?

Once I figure out how to get those Ecto telemetry events, and add some more of the Phoenix ones, then I will convert this to a real PR and start working towards an actual release.

**Comment by rabidpraxis** [comment]
@sorentwo A final look through would be awesome, it hasn't changed a ton since the last review.

I have a question about supporting multiple versions of dependencies in libraries. 

In a Ruby library, I would declare a minimum version for a dependency, and then if the API differs between versions, I can inspect that during runtime and massage the data to fit all situations. It is up to me to make sure I have all the permutations covered, but it can be nice for the consumer since they don't have to worry about compatibility. I tried to do this with telemetry, only requiring > 0.2.0, and trying to take care of the difference between using `Telemetry` (up to 0.3.0) vs `:telemetry` (after 0.3.0). I tried switching based on which application was loaded `Code.ensure_compiled?` but one branch always produced a warning, or failed on dialyzer.

Is this a thing that is supported at all? Do I just need to specify the latest dependency version, requiring users of the library to upgrade their deps?

File: mix.exs [hunk 7]
```diff
@@ -43,6 +43,7 @@ defmodule Honeybadger.Mixfile do
         api_key: {:system, "HONEYBADGER_API_KEY"},
         app: nil,
         breadcrumbs_enabled: false,
+        ecto_repos: nil,
```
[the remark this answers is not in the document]
rabidpraxis: I like it!

PR #254, opened 2020-01-28
Sanitize invalid binaries when encoding JSON

sorentwo:
Jason refuses to encode invalid JSON binaries, which results in an `{:error, %Jason.EncodeError{}}` result during encoding. We now match on encoding errors and attempt to coerce results into a valid binary.

### DIFF
--- a/lib/honeybadger/json.ex
+++ b/lib/honeybadger/json.ex
@@ -2,15 +2,14 @@ defmodule Honeybadger.JSON do
   @moduledoc false
 
   @spec encode(term) :: {:ok, String.t()} | {:error, Jason.EncodeError.t()}
-  def encode(notice) do
-    # try to encode without going the rabbit hole
-    case safe_encode(notice) do
+  def encode(term) do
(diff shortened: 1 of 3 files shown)


lib/honeybadger/json.ex
```diff
@@ -2,15 +2,19 @@ defmodule Honeybadger.JSON do
...
+  def encode(term) do
+    case safe_encode(term) do
       {:ok, output} ->
         {:ok, output}
 
+      {:error, %Jason.EncodeError{}} ->
+        term
+        |> to_encodeable()
+        |> Jason.encode()
+
```
rabidpraxis: Is there a reason to match on both of these explicit cases? Since they have the same result (and AFAICT these are the only possible errors) would it make sense to unify them into a single clause?

sorentwo: That's a fair point. Anything else should raise an exception here, so combining the cases is fine. I'll push up that change before merging.

PR #259, opened 2020-02-06
Elixir 1.10 fixes

rabidpraxis:
Added 2 fixes to address test failures in 1.10.

* Looks like some extra data has been added to the Logger metadata in 1.10, one of which includes a reference to a function, which was not included in our safe_encoding clauses. I added that.

* There are some tests asserting that Honeybadger.context will fail with a `FunctionClauseError` when the wrong type is passed. The guards that were in place were removed at one point, but the underlying implementation of `Logger.metadata/1` still returned the same error. Welp, the implementation was updated, so now it returns a Protocol error instead. I added the `is_list` guard back.

### DIFF
--- a/lib/honeybadger/json.ex
+++ b/lib/honeybadger/json.ex
@@ -43,7 +43,8 @@ defmodule Honeybadger.JSON do
     |> to_encodeable
   end
 
-  defp to_encodeable(input) when is_pid(input) or is_port(input) or is_reference(input) do
+  defp to_encodeable(input)
+       when is_pid(input) or is_port(input) or is_reference(input) or is_function(input) do
     inspect(input)
   end
 
--- a/test/honeybadger/json_test.exs
+++ b/test/honeybadger/json_test.exs
@@ -40,7 +40,7 @@ defmodule Honeybadger.JSONTest do
 
     test "handles values requring inspection" do
       {:ok, ~s("&Honeybadger.JSON.encode/1")} = JSON.encode(&Honeybadger.JSON.encode/1)
-      {:ok, ~s("#PID<0.250.0>")} = JSON.encode(:c.pid(0,250,0))
+      {:ok, ~s("#PID<0.250.0>")} = JSON.encode(:c.pid(0, 250, 0))
 
       ref = make_ref()
       {:ok, encoded_ref} = JSON.encode(ref)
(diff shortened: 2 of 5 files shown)


lib/honeybadger/json.ex
```diff
@@ -43,7 +43,8 @@ defmodule Honeybadger.JSON do
     |> to_encodeable
   end
 
-  defp to_encodeable(input) when is_pid(input) or is_port(input) or is_reference(input) do
+  defp to_encodeable(input)
```
sorentwo: It would be nice to have an example of this (any of these, really) in the json tests directly.

sorentwo:
@rabidpraxis @joshuap it would be great to get a release out after this is merged.

lib/honeybadger/json.ex
rabidpraxis: I added a test covering this to_encodable function

test/honeybadger/json_test.exs
```diff
@@ -38,6 +38,19 @@ defmodule Honeybadger.JSONTest do
       assert custom_encoded == jason_encoded
     end
 
+    test "handles values requring inspection" do
+      {:ok, ~s("&Honeybadger.JSON.encode/1")} = JSON.encode(&Honeybadger.JSON.encode/1)
+      {:ok, ~s("#PID<0.250.0>")} = JSON.encode(:c.pid(0,250,0))
```
sorentwo:
```suggestion
      {:ok, ~s("#PID<0.250.0>")} = JSON.encode(pid("0.250.0"))
```

sorentwo:
```suggestion
      {:ok, ~s("#PID<0.250.0>")} = JSON.encode(pid("0.250.0"))
```

sorentwo:
Well those comments got crazy. It won't let me commit the suggestion 😞

rabidpraxis:
@sorentwo I tried changing this but locally when I replace the Erlang pid function with the Elixir version I get:

`test/honeybadger/json_test.exs:43: undefined function pid/1`

sorentwo:
Doh! Sorry, that is `iex` only!

rabidpraxis:
Ah, explains why it worked in IEx.. If you are good with this, I can merge and work on a `0.13.1` release.

sorentwo:
Looks great to me 👍

PR #262, opened 2020-02-12
Handle json encoding Protocol.UndefinedError

### LINKED ISSUE #261: Client has unhandled encoding errors
[handle_cast notice](https://github.com/honeybadger-io/honeybadger-elixir/blob/master/lib/honeybadger/client.ex#L131) does not handle `%Protocol.UndefinedError`, which is the result I get for unencodable notices.  Enough of these errors can lead to the honeybadger app crashing and taking down the main application.

```
Honeybadger.JSON.encode(%{ test: fn () -> :noop end })
{:error,
 %Protocol.UndefinedError{
   description: "Jason.Encoder protocol must always be explicitly implemented",
   protocol: Jason.Encoder,
   value: #Function<20.99386804/0 in :erl_eval.expr/5>
 }}
```

kennethito:
Fixes #261

### DIFF
--- a/lib/honeybadger/client.ex
+++ b/lib/honeybadger/client.ex
@@ -130,6 +130,9 @@ defmodule Honeybadger.Client do
 
       {:error, %Jason.EncodeError{message: message}} ->
         Logger.warn(fn -> "[Honeybadger] Notice encoding failed: #{message}" end)
+
+      {:error, %Protocol.UndefinedError{description: message}} ->
+        Logger.warn(fn -> "[Honeybadger] Notice encoding failed: #{message}" end)
     end
 
     {:noreply, state}
--- a/lib/honeybadger/json.ex
+++ b/lib/honeybadger/json.ex
@@ -1,7 +1,10 @@
 defmodule Honeybadger.JSON do
   @moduledoc false
 
-  @spec encode(term) :: {:ok, String.t()} | {:error, Jason.EncodeError.t()}
+  @spec encode(term) ::
+          {:ok, String.t()}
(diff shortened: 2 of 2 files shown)


lib/honeybadger/client.ex
```diff
@@ -130,6 +130,9 @@ defmodule Honeybadger.Client do
 
       {:error, %Jason.EncodeError{message: message}} ->
         Logger.warn(fn -> "[Honeybadger] Notice encoding failed: #{message}" end)
+
+      {:error, %Protocol.UndefinedError{description: message}} ->
```
sorentwo: This is failing dialyzer because the structure seems to be wrong.

```suggestion
      {:error, %Protocol.UndefinedError{message: message}} ->
```

kennethito:
```
lib/honeybadger/client.ex:134: 
The pattern {'error', #{'description':=_message@2, '__struct__':='Elixir.Protocol.UndefinedError'}} 
can never match the type {
'error',#{'__exception__':=_, '__struct__':='Elixir.Jason.EncodeError', 'message':=binary()}} | {'ok',binary()}
```

The [Protocol.UndefinedError](https://github.com/elixir-lang/elixir/blob/5d57c9117494a97a6dd972d2334f19892aa0fafe/lib/elixir/lib/exception.ex#L1167) has description and not message.

I think the dialyzer error refers this [spec](https://github.com/honeybadger-io/honeybadger-elixir/blob/master/lib/honeybadger/json.ex#L4), which needs to be updated to return both errors

sorentwo:
Thanks @kennethito!

PR #265, opened 2020-02-19
Handle more edge cases in JSON encoding

moxley:
The JSON encoder tried to blacklist data types that needed to wrapped in `Kernel.inspect/1`, but that left at least one type (function) able to pass straight through to `Jason.encode/1`. That caused the entire Erlang VM to crash in same cases.

The commit changes the encoding logic to whitelist types that can be passed directly to `Jason.encode/1`, while wrapping everything else in `Kernel.inspect/1`.


moxley:
Closed, because I had an old version of master checked out, and this fix is already in the latest.

PR #294, opened 2020-06-09
Implement Jason.Encoder for RingBuffer

rabidpraxis:
We use logger metadata as a storage area for intermediate Breadcrumbs. Before we serialize and send a notice, we extract the RingBuffer and transform it into its final Map form. We do all this assuming that nobody will look into our special logger metadata. Welp, we have had a report of another library trying to output the logger metadata (and failing to serialize).

This PR implements `Jason.Encoder` for our RingBuffer struct, extracting and outputting an array of breadcrumbs (or whatever was stored in the RingBuffer).

### DIFF
--- a/test/honeybadger/breadcrumbs/ring_buffer_test.exs
+++ b/test/honeybadger/breadcrumbs/ring_buffer_test.exs
@@ -18,4 +18,13 @@ defmodule Honeybadger.Breadcrumbs.RingBufferTest do
 
     assert buffer == [:b, :c]
   end
+
+  test "implements Jason.Encoder" do
+    json =
+      RingBuffer.new(2)
+      |> RingBuffer.add(123)
+      |> Jason.encode!()
+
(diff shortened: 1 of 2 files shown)


test/honeybadger/breadcrumbs/ring_buffer_test.exs
```diff
@@ -18,4 +18,13 @@ defmodule Honeybadger.Breadcrumbs.RingBufferTest do
 
     assert buffer == [:b, :c]
   end
+
+  test "implements Jason.Encoder" do
+    json =
+      RingBuffer.new(2)
+      |> RingBuffer.add(123)
+      |> Jason.encode!()
```
sorentwo, line 26: This is something Credo normally points out as a "raw pipe chain":

```suggestion
      2
      |> RingBuffer.new()
      |> RingBuffer.add(123)
      |> Jason.encode!()
```

Issue #233, opened 2019-07-23
[Feature] Add an option to define a fingerprint


axelson:
The ruby library allows you to specify the `:fingerprint` which controls the grouping. It would be nice to be able to control the fingerprint in Elixir as well.

stympy:
The documentation for the Rayburn client for sending additional info along with the error is [here](https://docs.honeybadger.io/lib/ruby/getting-started/reporting-errors.html#passing-additional-options-to-honeybadger-notify). My thought is that we can do something similar as a solution for this issue. What do you think, @joshuap?

joshuap:
I think we could do something similar, although it may look different from Ruby. `Honeybadger.notify` in Elixir accepts an exact number of arguments, and I don't *believe* the API currently supports keyword arguments like we do in Ruby, such as `Honeybadger.notify(exception, fingerprint: 'string')`. Something like that might be a nice optimization if possible.

The other/additional way that we could support fingerprints in Elixir is to add a global config option which allows you to define a function that given a notice, returns the fingerprint for the notice (we have a similar option in Ruby, which uses a `Proc`).

I would say let's definitely add the global config, and also support the local/`Honeybadger.notify` option if feasible.

Any thoughts on the best implementation for this, @sorentwo?

sorentwo:
#### Global Fingerprints

@joshuap Are you imagining that the top level function would replicate the [before_notice](https://docs.honeybadger.io/lib/ruby/getting-started/customizing-error-grouping.html#customizing-the-grouping-for-all-exceptions) behavior? Or, would it be more localized and just be a function to calculate the fingerprint? Are there any other use cases for the `before_notify` callback?

#### Localized Fingerprints

The top level `Honeybadger.notify` function has multiple arities. Due to default arguments for both `metadata` and `stacktrace` there are three functions defined:

1. `notify(exception)` -> `notify(exception, %{}, [])`
2. `notify(exception, context)` -> `notify(exception, context, [])`
3. `notify(exception, context, stacktrace)` -> `notify(exception, context, stacktrace)`.

The best implementation I can think of is to define a new clause for `notify/2` with a signature like:

```elixir
notify(exception :: Notice.noticeable(), options :: Keyword.t())
```

Here we would accept `context`, `stacktrace`, `fingerprint`, etc. as options. We can use guards to keep the old `notify/2` where the second argument is a map and add this new more compact clause.

[...]

PR #305, opened 2020-08-02
Fingerprint option

brunozrk:
Implementation for fingerprint option, as discussed in https://github.com/honeybadger-io/honeybadger-elixir/issues/233

- new `notify!/2` option that accepts `metadata, stacktrace and fingerprint`
```
# we can't add `notify!/2` and keep the old because it conflicts the defaults
(CompileError) lib/honeybadger.ex:267: def notify/2 conflicts with defaults from notify/3

# the new has keyword as default in second param
def notify(exception, options \\ [])

# the old has map as default in second param
def notify(exception, metadata \\ %{}, stacktrace \\ [])
```

- fingerprint adapter behaviour that receives a `Noticiable` and must return a string, the precedence is as follow:
  - fingerprint option in `notify!/2`
  - fingerprint adapter

### DIFF
--- a/lib/honeybadger.ex
+++ b/lib/honeybadger.ex
@@ -41,7 +41,12 @@ defmodule Honeybadger do
         exception ->
           context = %{user_id: 1, account: "A Very Important Customer"}
 
-          Honeybadger.notify(exception, metadata: context, stacktrace: __STACKTRACE__, fingerprint: "user-1")
+          Honeybadger.notify(
+            exception,
+            metadata: context,
+            stacktrace: __STACKTRACE__,
+            fingerprint: "user-1"
+          )
       end
 
   Note that `notify` may be used outside of `try`, but it will use a different
--- a/lib/honeybadger.ex
+++ b/lib/honeybadger.ex
@@ -256,13 +256,13 @@ defmodule Honeybadger do
         }
 
   @spec notify(Notice.noticeable()) :: :ok
-  @spec notify(Notice.noticeable(), notify_options) :: :ok
   def notify(exception) do
     notify(exception, [])
   end
 
+  @spec notify(Notice.noticeable(), notify_options) :: :ok
   def notify(exception, metadata) when is_map(metadata) do
-    Logger.warn(
+    IO.warn(
       "Passing a metadata map is deprecated, " <>
         "use Honeybadger.notify(exception, metadata: metadata) instead"
     )
@@ -294,8 +294,10 @@ defmodule Honeybadger do
     |> Client.send_notice()
   end
 
+  @doc deprecated: "Use Honeybadger.notify/2 instead"
+  @spec notify(Notice.noticeable(), map(), list()) :: :ok
   def notify(exception, metadata, stacktrace) when is_map(metadata) and is_list(stacktrace) do
-    Logger.warn("Reporting with notify/3 is deprecated, use notify/2 instead")
+    IO.warn("Reporting with notify/3 is deprecated, use notify/2 instead")
     notify(exception, metadata: metadata, stacktrace: stacktrace)
   end
 
--- a/lib/honeybadger.ex
+++ b/lib/honeybadger.ex
@@ -251,7 +251,7 @@ defmodule Honeybadger do
 
   @type notify_options :: %{
           metadata: map(),
-          stacktrace: list(),
+          stacktrace: Exception.stacktrace_entry(),
           fingerprint: String.t()
         }
 
--- a/lib/honeybadger.ex
+++ b/lib/honeybadger.ex
@@ -160,6 +160,8 @@ defmodule Honeybadger do
 
   use Application
 
+  require Logger
+
   alias Honeybadger.{Client, Notice}
   alias Honeybadger.Breadcrumbs.{Collector, Breadcrumb}
 
@@ -241,8 +243,20 @@ defmodule Honeybadger do
       iex> Honeybadger.notify(%RuntimeError{}, %{culprit_id: 123})
       :ok
   """
-  @spec notify(Notice.noticeable(), map(), list()) :: :ok
-  def notify(exception, metadata \\ %{}, stacktrace \\ []) do
+  # @spec notify(Notice.noticeable(), map(), list()) :: :ok
+  def notify(exception, metadata \\ %{}, stacktrace \\ []) when is_map(metadata) do
+    Logger.warn(
+      "Honeybadger.notify/1, Honeybadger.notify/2 and Honeybadger.notify/3 are deprecated, please use Honeybadger.notify!/2 instead"
+    )
+
+    notify!(exception, metadata: metadata, stacktrace: stacktrace)
+  end
+
+  def notify!(exception, options \\ []) do
+    metadata = options[:metadata] || %{}
+    stacktrace = options[:stacktrace] || []
+    fingerprint = options[:fingerprint] || ""
+
     # Grab process local breadcrumbs if not passed with call and add notice breadcrumb
     breadcrumbs =
       metadata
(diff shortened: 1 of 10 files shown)


brunozrk:
Thanks for the review @sorentwo.

Actually the reason I created a new `notify!` function is because it is not possible to have different defaults for the same second parameter:

current `notify` method, a map:

```
def notify(exception, metadata \\ %{}, stacktrace \\ [])
```

the new (without bang), a keyword:

```
def notify(exception, options \\ [])
```

It conflicts the defaults, we get a compile error:
```
def notify/2 conflicts with defaults from notify/3
```

brunozrk:
to keep the same name, we could do something like that, was that your point?

```
  def notify(exception, options \\ %{}, stacktrace \\ []) do
    cond do
     is_map(options) -> do_notify(exception, metadata: options, stacktrace: stacktrace) # here we can wanr deprecation
      Keyword.keyword?(options) -> do_notify(exception, options)
    end
  end

  defp do_notify(exception, options) do
    ...
  end
```

sorentwo:
Right, you can't do this with the default argument syntax. It requires individual clauses for each `notify/1`, `notify/2` and `notify/3`. The benefit of individual clauses is that you can have custom deprecations for `notify/1` and `notify/3` though. 

I took a pass at writing out the clauses with warnings and deprecations (though as is this will definitely break some other internal code):

```elixir
  @doc deprecated: "Use notify/2 instead"
  def notify(exception) do
    IO.warn("Reporting with notify/1 is deprecated, use notify/2 instead")

    notify(exception, [])
  end

  @doc """
  Put the normal docs here...
  """
  def notify(exception, %{breadcrumbs: breadcrumbs} = metadata) do
    IO.warn("Passing breadcrumbs in the metadata map is deprecated, " <> "use [breadcrumbs: breadcrumbs] instead")

    metadata = Map.drop(metadata, :breadcrumbs)

    notify(exception, breadcrumbs: breadcrumbs, metadata: metadata)
  end

  def notify(exception, metadata) when is_map(metadata) do
    IO.warn("Passing a metadata map to notify/2 is deprecated, use [context: context] instead")

    notify(exception, metadata: metadata)
  end

  def notify(exception, options) when is_list(options) do
    breadcrumbs =
      options
      |> Keyword.get(:breadcrumbs, Collector.breadcrumbs())
      |> Collector.put(notice_breadcrumb(exception))
      |> Collector.output()

    options =
      options
      |> Keyword.put(:breadcrumbs, breadcrumbs)
      |> Keyword.put_new(:context, %{})
      |> Keyword.put_new(:stacktrace, [])

    exception
    |> Notice.new(options)
    |> Client.send_notice()
  end

  @doc deprecated: "Use notify/2 instead"
  def notify(exception, metadata, stacktrace) when is_map(metadata) and is_list(stacktrace) do
    IO.warn("Reporting with notify/3 is deprecated, use notify/2 instead")

    notify(exception, metadata: metadata, stacktrace: stacktrace)
  end
```

brunozrk:
got it, just updated the code, thx @sorentwo

brunozrk:
gentle ping @sorentwo

lib/honeybadger.ex
```diff
@@ -238,11 +240,41 @@ defmodule Honeybadger do
...
+
+      iex> Honeybadger.notify(%RuntimeError{}, fingerprint: "culprit_id-123")
       :ok
   """
-  @spec notify(Notice.noticeable(), map(), list()) :: :ok
-  def notify(exception, metadata \\ %{}, stacktrace \\ []) do
+
+  @type notify_options :: %{
+          metadata: map(),
+          stacktrace: list(),
```
sorentwo, line 254:
```suggestion
          stacktrace: Exception.stacktrace_entry(),
```

lib/honeybadger.ex
```diff
@@ -238,11 +240,41 @@ defmodule Honeybadger do
...
+          fingerprint: String.t()
+        }
+
+  @spec notify(Notice.noticeable()) :: :ok
+  @spec notify(Notice.noticeable(), notify_options) :: :ok
```
sorentwo, line 259: Can we move this down to where `notify/2` is defined?

lib/honeybadger.ex
```diff
@@ -257,10 +289,32 @@ defmodule Honeybadger do
...
 
     exception
-    |> Notice.new(metadata_with_breadcrumbs, stacktrace)
+    |> Notice.new(metadata_with_breadcrumbs, stacktrace, fingerprint)
+    |> put_notice_fingerprint()
     |> Client.send_notice()
   end
 
+  def notify(exception, metadata, stacktrace) when is_map(metadata) and is_list(stacktrace) do
+    Logger.warn("Reporting with notify/3 is deprecated, use notify/2 instead")
```
sorentwo, line 298: Using `Logger.warn` is different than `IO.warn` as it isn't caught by the compiler or `--warnings-as-errors`. To properly deprecate this we'll need to use `IO.warn` instead.

lib/honeybadger.ex
```diff
@@ -257,10 +289,32 @@ defmodule Honeybadger do
       |> Map.put(:breadcrumbs, breadcrumbs)
 
     exception
-    |> Notice.new(metadata_with_breadcrumbs, stacktrace)
+    |> Notice.new(metadata_with_breadcrumbs, stacktrace, fingerprint)
+    |> put_notice_fingerprint()
     |> Client.send_notice()
   end
 
+  def notify(exception, metadata, stacktrace) when is_map(metadata) and is_list(stacktrace) do
```
sorentwo, line 297: This still needs a spec, and we can add a `@doc deprecated: "Use Honeybadger.notify/2 instead"` to point people in the right direction from the docs.

lib/honeybadger/fingerprint_adapter.ex
```diff
@@ -0,0 +1,7 @@
+defmodule Honeybadger.FingerprintAdapter do
+  @moduledoc """
+  The callbacks required to implement the FingerprintAdapter behaviour
+  """
+
+  @callback parse(Notice.noticeable()) :: String.t()
```
sorentwo, line 6: It would be nice to document the callback itself as well. Possibly with some examples of returning a string based on the notice itself, examples of what a fingerprint should look like, etc.

brunozrk:
Code updated 👍

lib/honeybadger.ex
```diff
@@ -41,7 +41,7 @@ defmodule Honeybadger do
         exception ->
           context = %{user_id: 1, account: "A Very Important Customer"}
 
-          Honeybadger.notify(exception, context, __STACKTRACE__)
+          Honeybadger.notify(exception, metadata: context, stacktrace: __STACKTRACE__, fingerprint: "user-1")
```
sorentwo, line 44:
```suggestion
          Honeybadger.notify(
            exception,
            metadata: context,
            stacktrace: __STACKTRACE__,
            fingerprint: "user-1"
          )
```

lib/honeybadger/fingerprint_adapter.ex
```diff
@@ -0,0 +1,16 @@
+defmodule Honeybadger.FingerprintAdapter do
+  @moduledoc """
+  The callbacks required to implement the FingerprintAdapter behaviour
+  """
+
+  @doc """
+  For applications that specifies a fingerprint_adapter. This function receives
+  a Notice and must return a string that will be used as a fingerprint for the
+  request:
```
sorentwo, line 9:
```suggestion
  This function receives a `t:Notice.t/0` and must return a string that will be used as a
  fingerprint for the request:
```

sorentwo (approved):

sorentwo:
Thanks for all your hard work @brunozrk! Very nicely done 💛

Issue #276, opened 2020-04-01
Notify on :error level Logger messages


rabidpraxis:
We had a customer that expected the `Logger.error` call to notify Honeybadger. It looks like it would be a trivial change to allow notifying on logger error calls.

My question (@joshuap @sorentwo) is there a reason that we only log on process crashes?

sorentwo:
There isn't any reason not to do this now. We already use the error logger to do our reporting. Slight refinement though, we notify on both exceptions (rescue) and crashes (catch).

yukster:
Hey, just commenting here instead of #315 because this one is still open. I question whether this was a sensible choice. I work on several large and long-running Elixir apps and was surprised to see my HB errors go through the roof after updating a bunch of deps, including the HB client. 

I'm going to unhook it from the logger by setting `use_logger` to false (and hoping that will *only* turn off the Logger.error calls going to HB) but it seems like assuming calling Logger.error is an exceptional case is a bit presumptuous. We log inter-service communication issues at error level and want to track their volume (in Datadog) but we don't need those in HB. I feel like Honeybadger is a triage system for unexpected errors: something new shows up after a deploy, you make a story (we have the Jira plugin), work up a fix, deploy it, and the error is resolved. But anything that goes into Honeybadger due to a Logger call is obviously never going to be resolved!

Anyway, my two cents... And probably a moot argument if the default for use_logger is false... someone please holler if setting `use_logger` to false will affect anything other than stopping Logger.error calls from notifying HB!

yukster:
@joshuap thanks for the fast response! I have to say, I'm a bit confused by the SASL logging docs, especially because it says that SASL logger was deprecated in OTP 21 (I think we're on 23 across the board). Well, I'm going to test this throughly to make sure crashes still go to HB.

yukster:
So I confirmed that I will only get notifications from the plug integration with `use_logger: false`... so no error-reporting for GenServers and the like. I think I'm just going to roll back to 0.15.0 for now. And I would love to see a config for logging errors. Maybe I can whip up a PR for it.

TraceyOnim:
> @yukster good points.
> 
> I _think_ that `use_logger: false` will also [prevent reporting crashes in SASL compliant processes](https://github.com/honeybadger-io/honeybadger-elixir#logger).
> 
[...]

PR #315, opened 2020-10-02
Send notifications on error-level logs

saneery:
👋 
Relates to https://github.com/honeybadger-io/honeybadger-elixir/issues/276
I set the exception as `RuntimeError`, otherwise, the message will be with the prefix `Erlang Error:`. 

Should I add a configuration to disable and enable notifications?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ CHANGELOG](http://keepachangelog.com/) for how to update this file. This project
 adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
+### Changed
+- Send notifications on error-level logs
 
 ## [v0.15.0] - 2020-09-24
 ### Added
--- a/lib/honeybadger/logger.ex
+++ b/lib/honeybadger/logger.ex
@@ -40,7 +40,7 @@ defmodule Honeybadger.Logger do
         notify(reason, full_context, [])
 
       _ ->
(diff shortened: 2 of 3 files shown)


sorentwo (commented):
The changes in here look great to me. I'm wondering if there should be a configuration option to disable this behavior, since it is a potential source of many new notices. Thoughts @joshuap?

saneery:
Name `include_error_logs` is ok?

joshuap:
I'm ambivalent about having an option--I'd be OK if we ship this without one and then add an option if people request the ability to disable this type of logging specifically (we already have the `user_logger` option, though, right?)

sorentwo:
@joshuap That works for me 👍

saneery:
@sorentwo could you, please, add `hactoberfest-accepted` label to this PR? 🙏 
here's the instruction https://hacktoberfest.digitalocean.com/hacktoberfest-update

sorentwo:
I'll defer to @joshuap on the hacktoberfest label!

joshuap:
@saneery sorry, I was on vacation for two weeks and missed this! I added the label. I hope it still counts. :) Thanks for the PR! 🏆

saneery:
> @saneery sorry, I was on vacation for two weeks and missed this! I added the label. I hope it still counts. :) Thanks for the PR! 🏆

no problem, it was counted without the label :)

PR #334, opened 2021-02-13
Ignored domains

### LINKED ISSUE #242: Multiple errors reporting
We have had a few reports of 2 errors reporting. Based on the notice data, it seems like both the `Plug` error handler and the logger are reporting the same exception. It appears something like this has been a problem before (#52, #56) and was resolved.

I  have noticed this when an error is thrown in the an Ecto.Repo process (say an invalid query) then I see an error reported for the Repo process and also the Phoenix/Plug process.

rabidpraxis:
Should resolve #242 

This change allows for ignoring specific logger error events based on the log event domain metadata. Using the supplied defaults, this should deal with reported duplicate errors between Honeybadger.Plug and Honeybadger.Logger.

The logger will now ignore any events that are of the :cowboy domain by default. Since Phoenix integrates with cowboy to execute plugs, the cowboy error is often thrown at the same path as our Plug.ErrorHandler.

It's been a while since I have written any Elixir, so there might be better ways to do this. I'm also unsure if the method of pulling from the global config is best practice either.

I still need to add docs. I will do that next week.

### DIFF
--- a/lib/honeybadger/logger.ex
+++ b/lib/honeybadger/logger.ex
@@ -11,9 +11,7 @@ defmodule Honeybadger.Logger do
   end
 
   def init({__MODULE__, opts}) when is_list(opts) do
-    level = Keyword.get(opts, :level)
-
-    {:ok, %{level: level}}
+    {:ok, %{ level: Keyword.get(opts, :level) }}
   end
 
   @impl true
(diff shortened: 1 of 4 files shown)


lib/honeybadger/logger.ex
```diff
@@ -11,9 +11,7 @@ defmodule Honeybadger.Logger do
   end
 
   def init({__MODULE__, opts}) when is_list(opts) do
-    level = Keyword.get(opts, :level)
-
-    {:ok, %{level: level}}
+    {:ok, %{ level: Keyword.get(opts, :level) }}
```
sorentwo, line 14:
```suggestion
    {:ok, %{level: opts[:level]}}
```

sorentwo (approved):
Nice solution to the problem! One tiny nitpick, no blockers 👍

rabidpraxis:
Thanks for the review @sorentwo!

Issue #308, opened 2020-08-17
Is there a way to ignore an entire exception?


kennethito:
I see that you can filter out parts of an exception, but I was wondering if there was a way to selectively ignore and not notify an entire exception.

Use case.  The version of cowboy we are on generates an exception for certain characters in urls via `:cow_uri.urldecode/2
`. We've had millions of these in the last week due to people probing our site and blew through our honeybadger plan.  I'm looking for a way to ignore or rate limit these exceptions.

joshuap:
@kennethito you can ignore the error in the Honeybadger UI and we'll stop alerting you about it, but I don't think we have a way to completely ignore the error in the client, which would be better (since then it wouldn't be sent to our servers at all). @sorentwo we don't currently have a solution for this that I am missing, do we?

kennethito:
Gotcha.

@joshuap Ignoring the error in the Honeybadger UI doesn't prevent the error from consuming you're quota correct?

sorentwo:
> @sorentwo we don't currently have a solution for this that I am missing, do we?

No, there isn't any way to declare that certain exceptions should be ignored on the client side currently.

joshuap:
> Ignoring the error in the Honeybadger UI doesn't prevent the error from consuming your quota correct?

Correct, they do not consume quota!

In any case, I think this is a feature we want to eventually add, so I'll leave this open to track it.

sezaru:
Is this still a wanted feature? If I create a PR to support filtering out exceptions would it be considered to be merged?

joshuap:
@sezaru yeah, I would like this feature. What kind of configuration API are you thinking? @sorentwo, do you have any thoughts on the best approach?

Here's how we do this in Ruby and JavaScript:

https://docs.honeybadger.io/lib/ruby/getting-started/ignoring-errors.html
https://docs.honeybadger.io/lib/javascript/guides/reducing-noise.html

sezaru:
I was thinking about a very simple behaviour with a single `filter/1` function that receives the `Notify` struct and the user can define its own logic to when and how he wants to filter the exception.

Thanks for the links, I just took a fast look at it and it seems that they allow more complex filter cases like by exception type, etc.

For my use case, only the behaviour described above is enough, but if it is desirable, I could simply have some already implemented adapters for that behaviour for each use case.

For example, an adapter that reads a config with the exception types they want to ignore and filters then, etc.

Do you have any better suggestion or something like that would do?

[...]

PR #342, opened 2021-03-07
Adds config option exclude_errors

sezaru:
This PR adds support to exclude errors from being sent to the Honeybadger servers. This was discussed in issue #308.

My approach was to make this as simple as possible and not opinative, giving the full responsibility and power to the user to exclude errors in their desired way.

This is done by implementing the `Honeybadger.ExcludeErrors` behaviour function `exclude_error?/1`, this function receives the full `Honeybadger.Notice` and should return a boolean signalling the error exclusion or not.

By default, the library will not exclude any errors from being sent.

Note that tests are missing because I first want to know that the changes are ok to you guys.

Also note that this can be easily expanded in the future to ship with more exclusion implementations out-of-box, as an example, I will post bellow my full implementation for a rate-limit exclusion strategy where I limit the number of messages per error that can be sent in a time window:

*exclude_errors.ex:*
```elixir
defmodule Log.Honeybadger.ExcludeErrors do
  alias Log.Honeybadger.ExcludeErrors.Server

  alias Honeybadger.ExcludeErrors

  @behaviour ExcludeErrors

  @spec child_spec(keyword) :: Supervisor.child_spec()
  defdelegate child_spec(opts), to: Server

  @impl ExcludeErrors
  def exclude_error?(notice), do: GenServer.call(Server, {:exclude_error?, notice}, :infinity)
end
```

*exclude_errors/server.ex:*
```elixir
defmodule Log.Honeybadger.ExcludeErrors.Server do
  use GenServer

  alias Log.Honeybadger.ExcludeErrors.Impl

  def start_link(_), do: GenServer.start_link(__MODULE__, :noargs, name: __MODULE__)

  @impl GenServer
  def init(_), do: {:ok, Impl.initialize()}

  @impl GenServer
  def handle_call({:exclude_error?, notice}, _, state) do
    {reply, state} = Impl.exclude_error?(notice, state)

    {:reply, reply, state}
  end

  @impl GenServer
  def handle_info({:timeout, timer, :decrement_counters}, %{timer: timer} = state),
    do: {:noreply, Impl.decrement_counters(state)}

  def handle_info({:timeout, _, :decrement_counters}, state), do: {:noreply, state}
end
```

*exclude_errors/impl.ex:*
```elixir
defmodule Log.Honeybadger.ExcludeErrors.Impl do
  alias Log.Honeybadger.ExcludeErrors.State

  alias Honeybadger.Notice

  @spec initialize :: State.t()
  def initialize, do: State.new!(self())

  @spec exclude_error?(Notice.t(), State.t()) :: {boolean, State.t()}
  def exclude_error?(notice, state) do
    %{error: %{fingerprint: fingerprint, class: class, backtrace: backtrace}} = notice

    if fingerprint != "", do: exclude?(fingerprint, state), else: exclude?({class, backtrace}, state)
  end

  @spec decrement_counters(State.t()) :: State.t()
  def decrement_counters(state) do
    state
    |> State.all()
    |> Enum.reduce(state, fn {key, _}, state ->
      {current_errors, state} = State.decrement(state, key)

      if current_errors == 0, do: State.delete(state, key), else: state
    end)
    |> State.start_timer(self())
  end

  defp exclude?(key, %{max_errors: max_errors} = state) do
    {current_errors, state} = State.increment(state, key)

    {current_errors == max_errors, state}
  end
end
```

*exclude_errors/state.ex:*
```elixir
defmodule Log.Honeybadger.ExcludeErrors.State do
  alias Common.Timer

  use TypedStruct

  @max_errors Application.compile_env!(:honeybadger, :max_errors) + 1
  @decrement_period Application.compile_env!(:honeybadger, :decrement_period)

  typedstruct enforce: true do
    field :max_errors, integer, default: @max_errors
    field :counter, reference
    field :timer, reference, enforce: false
  end

  @spec new!(pid) :: t
  def new!(from) do
    counter = :ets.new(Module.concat(__MODULE__, Counter), [:set, :protected])

    struct!(__MODULE__, counter: counter) |> start_timer(from)
  end

  @spec increment(State.t(), term) :: {integer, State.t()}
  def increment(%{counter: table} = state, key) do
    current_errors = :ets.update_counter(table, key, {2, 1, 5, 5}, {key, 0})

    {current_errors, state}
  end

  @spec decrement(State.t(), term) :: {integer, State.t()}
  def decrement(%{counter: table} = state, key) do
    current_errors = :ets.update_counter(table, key, {2, -1, 0, 0}, {key, 0})

    {current_errors, state}
  end

  @spec delete(State.t(), term) :: State.t()
  def delete(%{counter: table} = state, key) do
    :ets.delete(table, key)

    state
  end

  @spec all(State.t()) :: [{term, integer}]
  def all(%{counter: table}), do: :ets.tab2list(table)

  @spec start_timer(State.t(), pid) :: State.t()
  def start_timer(state, from),
    do: %{state | timer: Timer.start(from, :decrement_counters, @decrement_period)}
end
```


sezaru:
CI build has failed, but it seems to be an issue with the CI itself and not my changes as far as I can tell.

lib/honeybadger.ex
```diff
@@ -286,10 +287,14 @@ defmodule Honeybadger do
...
-    |> put_notice_fingerprint()
-    |> Client.send_notice()
+    notice =
+      exception
+      |> Notice.new(metadata_with_breadcrumbs, stacktrace, fingerprint)
+      |> put_notice_fingerprint()
+
+    exclude_error_module = Application.get_env(:honeybadger, :exclude_errors)
+
+    if not exclude_error_module.exclude_error?(notice), do: Client.send_notice(notice)
```
TraceyOnim, line 297:
```suggestion
    unless exclude_error_module.exclude_error?(notice), do: Client.send_notice(notice)
```
Instead of negating within an `if .. else`, I think is better to use `unless`. What do you think about this?

TraceyOnim:
@sezaru If I override the default configuration to exclude errors, none of the exception will be sent. How is this supposed to work? 
Does a user has the power to selectively ignore an exception and not notify when they implement their own custom `Honeybadger.ExcludeErrors` behaviour?

sezaru:
Hi @TraceyOnim , I don't know if I understood your question correctly. do you mean that exceptions are handled differently inside Honeybadger? When I looked into the code I only found one way to send the errors.

> Does a user have the power to selectively ignore an exception and not notify when they implement their own custom Honeybadger.ExcludeErrors behaviour?

Assuming that the exceptions are handled in the same part that I changed in the code, the answer would be yes, the user would receive the full `Honeybadger.Notice` struct, analyze it and decide if it wants to send it or not.

TraceyOnim:
> Hi @TraceyOnim , I don't know if I understood your question correctly. do you mean that exceptions are handled differently inside Honeybadger? When I looked into the code I only found one way to send the errors.
> 
> > Does a user have the power to selectively ignore an exception and not notify when they implement their own custom Honeybadger.ExcludeErrors behaviour?
> 
> Assuming that the exceptions are handled in the same part that I changed in the code, the answer would be yes, the user would receive the full `Honeybadger.Notice` struct, analyze it and decide if it wants to send it or not.

Yeah  you got my question , it's now clear .Thank you for the clarification

jdubya80:
Is this something that is planned to be introduced at some point?  Is there another way to do "error exclusion"?

joshuap:
> Is this something that is planned to be introduced at some point? Is there another way to do "error exclusion"?

@jdubya80 there isn't another way, and I'm all for including this.

@sezaru do you want to add some tests for this? I think it looks good otherwise. I might want to extend this in the future so that the default implementation will ignore a list of error names if you configure something like this: `exclude_errors: [ArithmeticError, FunctionClauseError]`.

PR #353, opened 2021-04-02
Call custom filter_cgi_data/1 in NoticeFilter

### LINKED ISSUE #350: Custom filter_cgi_data/1 never called
I've been trying to write a custom filter to handle `cgi_data` a little differently than the default, and I noticed that my overridden `filter_cgi_data/1` was not being called. After a little poking around, I found that as of #105, the configured filter is [skipped for `cgi_data` in the default `NoticeFilter`](https://github.com/honeybadger-io/honeybadger-elixir/blob/v0.16.0/lib/honeybadger/notice_filter/default.ex#L21), with no way to override it.

I know I can write my own `NoticeFilter`, but that seems like overkill when I'm trying to do something using a specific, documented behaviour/callback.

joshuap:
I *think* this should fix the issue?

Fixes #350

### DIFF
--- a/lib/honeybadger/notice_filter/default.ex
+++ b/lib/honeybadger/notice_filter/default.ex
@@ -18,7 +18,7 @@ defmodule Honeybadger.NoticeFilter.Default do
     request
     |> apply_filter(:context, &filter.filter_context/1)
     |> apply_filter(:params, &filter.filter_params/1)
-    |> apply_filter(:cgi_data, &filter_cgi_data/1)
+    |> apply_filter(:cgi_data, &filter.filter_cgi_data/1)
     |> apply_filter(:session, &filter.filter_session/1)
     |> disable(:filter_disable_url, :url)
     |> disable(:filter_disable_session, :session)


lib/honeybadger/notice_filter/default.ex
```diff
@@ -19,6 +19,7 @@ defmodule Honeybadger.NoticeFilter.Default do
     |> apply_filter(:context, &filter.filter_context/1)
     |> apply_filter(:params, &filter.filter_params/1)
     |> apply_filter(:cgi_data, &filter_cgi_data/1)
+    |> apply_filter(:cgi_data, &filter.filter_cgi_data/1)
```
sorentwo, line 22: This would go in the filter_cgi_data function below. Oddly, the filter module isn't passed in there so it looks it up again.

joshuap: Ah, I see. I'm not sure what the correct fix is, then.

sorentwo: @joshuap I pushed an update to this branch that should fix it.

Issue #336, opened 2021-02-19
Errors shown as ErlangError instead of the specific error


sezaru:
I don't know if this is expected since it is the first time I used Honeybadger, but most of the errors I'm receiving are encapsulated inside `ErlangErrors`. For example this one:

![image](https://user-images.githubusercontent.com/279828/108510732-a3e41d80-729d-11eb-8cd5-23262e02513a.png)

As you can see, two of the errors are actually `Postgrex.Error` errors but they are encapsulated inside an `ErlangError` for some reason. This gives me 2 problems since I lose the nice stack trace view (the whole stack trace is contained inside the message part of the error view page.

And way worse this makes Honeybadger combine errors that don't have anything to do with one another as the same error. You can see this happening in the above image since it combined 2 websocket `:close` errors with 2 `Postgrex` errors.

Is there some option to fix this or is this a bug? I'm using the latest version and I also have `sasl` enabled with the option `config :logger, handle_sasl_reports: true`.

joshuap:
I'm not sure about this one. @sorentwo @rabidpraxis does this look familiar?

sorentwo:
@joshuap This is extremely common. Erlang errors are wrapped by Elixir and will all look like the same error. Some other fingerprinting is necessary to properly differentiate between them.

Having the Postgrex errors wrapped in an ErlangError is strange, though. That may be due to the `handle_sasl_reports: true` option, as I haven't encountered that particular issue (and dscout has **many** mis-grouped errors).

joshuap:
> This is extremely common. Erlang errors are wrapped by Elixir and will all look like the same error. Some other fingerprinting is necessary to properly differentiate between them.

Makes sense, thanks! Is the fingerprinting something we could think about handling on our end, or is that unique to each user and better handled via config? (Maybe related: #338)

sorentwo:
The `ErlangError` grouping is extremely common and something I'd expect to see handled by the server or library, not something that users should need to do themselves.

joshuap:
Ok, great. I'll keep this open as an enhancement issue... we'll see if we can find someone to work on that at some point.

sezaru:
I disabled `handle_sasl_reports` to see if it improve the situation, now I just need to wait for some errors to pop-up and I will update here my findings.

sezaru:
At first glance it seems to be working fine now with sasl turned off:
[...]

PR #359, opened 2021-04-16
Fix custom class and message notices

rktjmp:
It seemed that 

```
iex(__@__)4> Honeybadger.notify(%{class: "CustomClass", message: "a matching message"})
:ok
```

Were not being reported in the way they should:

image

Adds new function head to match the custom format. I can't imagine any "real" exception would ever match that pattern, they should all be structs, so ... probably safe?

See also: #336 

NB "class" is kind of a weird term (over maybe "type" or even just "name") in Elixir but I assume it's to keep the API uniform across languages.

### DIFF
--- a/lib/honeybadger/notice.ex
+++ b/lib/honeybadger/notice.ex
@@ -47,15 +47,27 @@ defmodule Honeybadger.Notice do
     new(%RuntimeError{message: message}, metadata, stacktrace, fingerprint)
   end
 
+  def new(%{class: exception_name, message: message}, metadata, stacktrace, fingerprint) do
+    do_new(exception_name, message, metadata, stacktrace, fingerprint)
+  end
+
   def new(exception, metadata, stacktrace, fingerprint)
       when is_map(metadata) and is_list(stacktrace) do
     {exception, stacktrace} = Exception.blame(:error, exception, stacktrace)
 
     %{__struct__: exception_mod} = exception
 
+    class = Utils.module_to_string(exception_mod)
+    message = exception_mod.message(exception)
+
+    do_new(class, message, metadata, stacktrace, fingerprint)
+  end
+
+  # bundles exception (or pseudo exception) information in to notice
+  defp do_new(class, message, metadata, stacktrace, fingerprint) do
     error = %{
-      class: Utils.module_to_string(exception_mod),
-      message: exception_mod.message(exception),
+      class: class,
+      message: message,
       backtrace: Backtrace.from_stacktrace(stacktrace),
       tags: Map.get(metadata, :tags, []),
       fingerprint: fingerprint
--- a/lib/honeybadger/notice.ex
+++ b/lib/honeybadger/notice.ex
@@ -49,7 +49,7 @@ defmodule Honeybadger.Notice do
 
   def new(%{class: exception_name, message: message}, metadata, stacktrace, fingerprint)
       when is_map(metadata) and is_list(stacktrace) do
-    do_new(exception_name, message, metadata, stacktrace, fingerprint)
+    new(exception_name, message, metadata, stacktrace, fingerprint)
   end
 
   def new(exception, metadata, stacktrace, fingerprint)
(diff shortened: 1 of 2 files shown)


lib/honeybadger/notice.ex
```diff
@@ -47,15 +47,27 @@ defmodule Honeybadger.Notice do
     new(%RuntimeError{message: message}, metadata, stacktrace, fingerprint)
   end
 
+  def new(%{class: exception_name, message: message}, metadata, stacktrace, fingerprint) do
+    do_new(exception_name, message, metadata, stacktrace, fingerprint)
```
sorentwo, line 51: Can we do avoid the `do_new` private function, modify the payload instead, and pass it to `new/4` again? We don't have much (any?) use of `do_` prefixed private functions elsewhere.

sorentwo (changes requested):
Nice catch and thanks for the PR. The test looks perfect, and the implementation itself is great. I'd like to rework it a little and avoid a `do_new` private function, though.

lib/honeybadger/notice.ex
rktjmp: Err... Pass it to which `new/4`?

I can't pass it as a string, seems against the point of passing a custom class and message.

I can't bundle it *back* into an exception or we are back where we started with ErlangError wrapping.

I could instead extract the exception into a `%{class: ..., message: ...}` call?

sorentwo: Excellent question. I was only looking at a subset of the clauses and didn't see the one before this. The whole `class` thing is strange anyhow!

We can't use a guard like `is_exception` because we support older Elixir versions. How about a private `new/5` function? It's what you wrote, but without the `do_prefix`.

rktjmp: Can do chief.

Should it also handle `%{class: xxx}`, and set a default message (even just a blank string)? Right now that will miss the match and end up as a ErlangError.

sorentwo: The only way to create one of these errors is by manually calling notify with a class payload, and we only document doing that with a `message` key as well. It's reasonable that sending an improper format causes an ErlangError. 

Honestly, we shouldn't encourage people to use the class/message variant and should only document sending a straight message or a proper exception.

lib/honeybadger/notice.ex
```diff
@@ -51,6 +51,12 @@ defmodule Honeybadger.Notice do
       when is_map(metadata) and is_list(stacktrace) do
     new(exception_name, message, metadata, stacktrace, fingerprint)
   end
+
+  def new(%{class: exception_name}, metadata, stacktrace, fingerprint)
+      when is_map(metadata) and is_list(stacktrace) do
+    new(%{class: exception_name, message: ""}, metadata, stacktrace, fingerprint)
```
rktjmp, line 57: @sorentwo I think it's better form to chain back to the "real map handler" over calling down to `new(exception_name, "", ...)` but change.

rktjmp: Nevermind, just saw comment. Will remove.

lib/honeybadger/notice.ex
rktjmp: My reason for wanting `%{class..., message: ...}` is paths that are a failure in some respect, I *do* want a notification about it, _but_ it isn't actually a runtime error.

In my case I may have a DB transaction fail. It's caught and handled but I don't want a specific error type for each transaction, I just want to know it went wrong and be able to tag them separately. So I need to be able to specify a ~~type~~ class for them, otherwise they call get bundled under "RuntimeError". Fingerprinting will separate them out but they're still ugly to parse in the UI.

I should just have a proper type but... well, passing some strings around can be a lot simpler.

Really I wish I could go `Honeybadger.notify("app.module_a.transaction_type", "failed: reason")`. Turn the "message" api into "title & message".

sorentwo (approved):
The change looks great now 👍 

Thanks for explaining your use case, that makes sense to me. It's pretty light weight to create custom errors and I suggest that you go that route. But, considering that the class/message format is already available, we need to support it cleanly.

PR #366, opened 2021-05-21
Fix iodata nil message issue

pthompson:
Fixes issue in Notice where IO.iodata_to_binary/1 would throw an ArgumentError exception if passed a nil message.


sorentwo (approved):
Nil isn't a particularly informative error message. Never the less, we don't want to error on what used to be valid input.

PR #369, opened 2021-06-28
Handle extra_info with error_info

abstractcoder:
The stacktrace `extra_info` field can contain an `error_info` tuple, along with the `file` and `line` tuples.

https://erlang.org/doc/reference_manual/errors.html#the-call-stack-back-trace--stacktrace-

My application was generating a stacktrace with an extra_info field that contained error_info. This was caused a FunctionClauseError exception when calling Honeybadger.Backtrace.format_line. This PR solves this issue.


joshuap (approved):
This looks good to me, thanks!

PR #370, opened 2021-06-30
Prevent internal error from causing infinite loop

### LINKED ISSUE #352: Rate limited response error causes infinite loop
I am seeing an infinite notification loop happen when hitting any exception reported by Honeybadger while having exceeded our account rate limit.

Upon reporting, the first response is `[warn] [Honeybadger] API failure: "{\"error\":\"Rate Limit Exceeded\"}`. It looks like this failure is then being caught at `client.ex:162`, but then `Logger.error` is called - sending this new notification and continuing the loop. This continues until the server is restarted.

## Proposed fixes

Switching the calls from `Logger.error` to `Logger.warn` stops the loop from starting. I'm happy to submit a PR with this but I'm not sure this is the ideal fix for everyone. 

An alternative might be to use @sezaru's exclude_error config option PR #342 but it would need to be merged in first.

Open to other solutions or ideas here! Thanks.

## Version info

Using honeybadger 0.16.0. 

Reproduced on multiple systems (Elixir 1.11.0-dev (82dca55) (compiled with Erlang/OTP 23) on MacOS 10.15.7) and (elixir 1.10.2 / Erlang 22.2 on Ubuntu 18.04.5 LTS).

Front logo [Front conversations](https://app.frontapp.com/open/top_3h2n5)

TraceyOnim:
This PR solves the issue  #352 . 
The problem is that when our API responds with an error code, and then we call `Logger. error`, the `use_logger `feature causes an additional error report to be sent, which causes the API to respond with another error, and it becomes an infinite cycle.
This has been done by first checking the Logger metadata received 
```
[
  erl_level: :error,
  application: :honeybadger,
  domain: [:elixir],
  file: "lib/honeybadger/client.ex",
  function: "post_notice/4",
  gl: #PID<0.465.0>,
  line: 181,
  mfa: {Honeybadger.Client, :post_notice, 4},
  module: Honeybadger.Client,
  pid: #PID<0.469.0>,
  time: 1625259235761950
]
```
If the application from metadata is `:honeybadger`,  the error notification is not sent to honeybadger API

Fixes #352 
Closes #372

### DIFF
--- a/lib/honeybadger/logger.ex
+++ b/lib/honeybadger/logger.ex
@@ -24,8 +24,9 @@ defmodule Honeybadger.Logger do
     {:ok, state}
   end
 
-  def handle_event({:error, _gl, {Logger, message, _ts, metadata}}, state) do
-    unless domain_ignored?(metadata[:domain], Honeybadger.get_env(:ignored_domains)) do
+  def handle_event({:error, _gl, {Logger, message, _ts, metadata}} = messo, state) do
+    unless domain_ignored?(metadata[:domain], Honeybadger.get_env(:ignored_domains)) ||
+             internal_error?(metadata) do
       details = extract_details(message)
       context = extract_context(metadata)
       full_context = Map.merge(details, context)
@@ -85,6 +86,13 @@ defmodule Honeybadger.Logger do
   end
 
   def domain_ignored?(_domain, _ignored), do: false
+
+  def internal_error?(metadata) do
+    case Keyword.get(metadata, :application) do
+      nil -> false
+      application -> application == :honeybadger
+    end
+  end
 
   @standard_metadata ~w(ancestors callers crash_reason file function line module pid)a
 
--- a/lib/honeybadger/logger.ex
+++ b/lib/honeybadger/logger.ex
@@ -26,7 +26,7 @@ defmodule Honeybadger.Logger do
 
   def handle_event({:error, _gl, {Logger, message, _ts, metadata}}, state) do
     unless domain_ignored?(metadata[:domain], Honeybadger.get_env(:ignored_domains)) ||
-             internal_error?(metadata) do
+             internal_error?(metadata[:application]) do
       details = extract_details(message)
       context = extract_context(metadata)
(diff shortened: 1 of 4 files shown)


sorentwo (approved):

joshuap (commented):
This looks great to me. I like it better than the approach using `ignored_domains` because it's impossible to override this by changing the configuration. Just a few final questions:

1. Is `application` always present when logging within the `Honeybadger` namespace?
2. Can you add a [changelog entry](https://github.com/honeybadger-io/honeybadger-elixir/blob/master/CHANGELOG.md#fixed)? The title and issue number of this PR would be fine.

lib/honeybadger/logger.ex
```diff
@@ -86,6 +87,13 @@ defmodule Honeybadger.Logger do
 
   def domain_ignored?(_domain, _ignored), do: false
 
+  def internal_error?(metadata) do
+    case Keyword.get(metadata, :application) do
+      nil -> false
+      application -> application == :honeybadger
+    end
+  end
```
joshuap, line 95: Could we eliminate the conditional by using pattern matching here?

```suggestion
def internal_error?(%{application: :honeybadger}), do: true
def internal_error?(_metadata), do: false
```

TraceyOnim:
> This looks great to me. I like it better than the approach using `ignored_domains` because it's impossible to override this by changing the configuration. Just a few final questions:
> 
> 1. Is `application` always present when logging within the `Honeybadger` namespace?
> 2. Can you add a [changelog entry](https://github.com/honeybadger-io/honeybadger-elixir/blob/master/CHANGELOG.md#fixed)? The title and issue number of this PR would be fine.

The application will always be present when honeybadger attempts to log an error.

Sure, l can add the changelog

lib/honeybadger/logger.ex
TraceyOnim: Working on it

joshuap (approved):
Great work. :shipit:

joshuap:
Released in 0.16.3

PR #372, opened 2021-07-04
Fix loop when the API server returns an error

luisgabrielroldan:
This PR stops the infinite loop when trying to post an error when the API returns an error.
The domain is used by the `Honeybadger.Client`post result logging.

The related issues are #352 and  #370.

### DIFF
--- a/lib/honeybadger/client.ex
+++ b/lib/honeybadger/client.ex
@@ -12,6 +12,7 @@ defmodule Honeybadger.Client do
   ]
   @max_connections 20
   @notices_endpoint "/v1/notices"
+  @post_logger_metadata [domain: [:honeybadger]]
 
   # State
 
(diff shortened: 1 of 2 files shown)


TraceyOnim:
I have pulled your PR and tested it using my pet project by adding the path of the cloned honeybadger project to the mix.exs `{: honeybadger, path: "~/honeybadger_projects/honeybadger-elixir"}` so that it can recognize the current changes. While Logging an error using the pet project I'm still getting an infinite loop. 
I might be doing something wrong, how did you test it?

luisgabrielroldan:
Sorry for that @TraceyOnim! I just fixed a mistake on the logger metadata.

joshuap:
@luisgabrielroldan thanks for the PR!! We ended up going with a slightly different solution (see #370), but hopefully with the same result. :) I've released the fix in 0.16.3—can you try that out and let me know if that fixes the issue for you?

luisgabrielroldan:
The issue is fixed! Thanks!

Issue #123, opened 2017-12-13
Improve argument reporting


joshuap:
Honeybadger now supports [sending arguments](https://github.com/honeybadger-io/honeybadger-elixir/pull/115), however the cases where the arguments are actually available seem limited; see [this comment by @sorentwo](https://github.com/honeybadger-io/honeybadger-elixir/pull/115#issuecomment-350154494).

We may be able to provide arguments more often by studying how Elixir/iex display the arguments in printed stack traces.

TraceyOnim:
I have found out, arguments are available when a FunctionClauseError is reported. Here is the [explanation](https://stackoverflow.com/questions/65406610/include-function-arguments-in-stack-trace-when-raising-error). 

With that, I have also found out that, the stack trace is been parsed at this [line](https://github.com/honeybadger-io/honeybadger-elixir/blob/master/lib/honeybadger/notice.ex#L57). Hence for the limited cases like FunctionClauseError, the arguments are not available for the final output.  There is an existing [PR](https://github.com/honeybadger-io/honeybadger-elixir/pull/new/fix/show_stacktrace_args) for the fix

joshuap:
@TraceyOnim nice!

So with #373 merged, what should the documentation say about `filter_args`? It still only works for `FunctionClauseError`, right? If that's the case, we should update the README with that information to make it clear how it works. Do you want to work on that?

Also, I added a changelog entry for #373: https://github.com/honeybadger-io/honeybadger-elixir/commit/44c15cf88a153cf45e12b9e341cf13febe7905fe

joshuap:
Another question: do you think it's possible to display args for other types of exceptions with some additional research/work in the future? Or should  I close this issue after the documentation is updated?

TraceyOnim:
> @TraceyOnim nice!
> 
> So with #373 merged, what should the documentation say about `filter_args`? It still only works for `FunctionClauseError`, right? If that's the case, we should update the README with that information to make it clear how it works. Do you want to work on that?
> 
> Also, I added a changelog entry for #373: [44c15cf](https://github.com/honeybadger-io/honeybadger-elixir/commit/44c15cf88a153cf45e12b9e341cf13febe7905fe)

From the research, I can say filter_args only works with FunctionClauseError.
Yes , I can work on the documentation

[...]

PR #373, opened 2021-07-09
show arguments in stacktrace

TraceyOnim:
This PR is related to #123 
 I have found out that, the stack trace is been parsed at this [notice.ex line 57](https://github.com/honeybadger-io/honeybadger-elixir/blob/master/lib/honeybadger/notice.ex#L57) Hence for the limited cases like FunctionClauseError, the arguments are not available for the final output. 
Here is the stack trace before  been parsed:
```
{HelloWeb.PageController, :match, [%{}],
   [file: 'lib/hello_web/controllers/page_controller.ex', line: 28]}
```
After been parsed at [notice.ex line 57](https://github.com/honeybadger-io/honeybadger-elixir/blob/master/lib/honeybadger/notice.ex#L57):
```
{HelloWeb.PageController, :match, 1,
   [file: 'lib/hello_web/controllers/page_controller.ex', line: 28]}
```

### DIFF
--- a/lib/honeybadger/notice.ex
+++ b/lib/honeybadger/notice.ex
@@ -54,7 +54,7 @@ defmodule Honeybadger.Notice do
 
   def new(exception, metadata, stacktrace, fingerprint)
       when is_map(metadata) and is_list(stacktrace) do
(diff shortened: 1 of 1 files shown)


sorentwo:
@TraceyOnim Nice find!

PR #390, opened 2021-08-24
update Elixir version in CI

TraceyOnim:


.github/workflows/elixir.yml
```diff
@@ -17,8 +17,8 @@ jobs:
       matrix:
         include:
           - pair:
-              elixir: '1.8'
-              otp: '21.3'
+              elixir: '1.12'
```
sorentwo, line 20: The general strategy is to list the oldest Elixir version we support along with the oldest Erlang version that works with it. In this case, I believe that means the first pair should be Elixir 1.10 / OTP 21.3, and then the next pair would be Elixir 1.12 / OTP 24.0

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ CHANGELOG](http://keepachangelog.com/) for how to update this file. This project
 adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
+### Changed
+- Update Elixir version in CI (#390)
```
sorentwo, line 9: We should be transparent about the new minimum required version (and the fact that the update is because we're tracking Phoenix).

sorentwo (changes requested):
Thanks for the changes. I've left a few notes to get the best CI coverage and to communicate the changes. One more thing, please bump the version in `mix.exs`, as that's the version used to resolve dependencies.

TraceyOnim:
> Thanks for the changes. I've left a few notes to get the best CI coverage and to communicate the changes. One more thing, please bump the version in `mix.exs`, as that's the version used to resolve

TraceyOnim:
> Thanks for the changes. I've left a few notes to get the best CI coverage and to communicate the changes. One more thing, please bump the version in `mix.exs`, as that's the version used to resolve dependencies.

Should I bump the version in master or in this branch?

.github/workflows/elixir.yml
TraceyOnim: Working on the changes, thank you for the clarification on that

CHANGELOG.md
TraceyOnim: I have made changes on that, kindly let me know if it's clear.

.github/workflows/elixir.yml
```diff
@@ -2,7 +2,7 @@ name: Elixir CI
 
 on:
   push:
-    branches: [ master ]
+    branches: [ master ]
```
sorentwo, line 5:
```suggestion
    branches: [ master ]
```

CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ CHANGELOG](http://keepachangelog.com/) for how to update this file. This project
 adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
+### Changed
+- Update Elixir version in CI (#390). Updating of phoenix dependancy from v1.5.10 to v1.5.11 resulted to a compilation error due to the tier dependancy(:mime) which requires Elixir 1.10+ but CI runs on Elixir v1.8.0 hence the need for an update to Elixir 1.10+.
```
sorentwo, line 9:
```suggestion
- Update the minimum Elixir version to 1.10 (#390). Phoenix v1.5.11 causes a compilation error due to the new `mime` dependency, which requires Elixir 1.10+.
```

sorentwo (approved):

CHANGELOG.md
sorentwo: Much better. I tweaked the language a bit on top of it 👍

PR #400, opened 2021-09-29
Avoid local function captures in Telemetry.attrach_many/4

0x6a68:
Local function captures should be avoided as they may cause performance penalty when calling such handler.
For more details see note in [Telemetry.attach_many/4](https://hexdocs.pm/telemetry/telemetry.html#attach_many/4) documentation.


sorentwo:
Excellent, thanks! 💛

sorentwo:
@joshuap Can we have a release? This causes a lot of noise still.

joshuap:
@sorentwo released in 0.18.1, thanks!

PR #442, opened 2022-08-02
Add Config Option exclude_error 

TraceyOnim:
Related to #443


lib/honeybadger.ex
```diff
@@ -287,10 +288,14 @@ defmodule Honeybadger do
...
-    exception
-    |> Notice.new(metadata_with_breadcrumbs, stacktrace, fingerprint)
-    |> put_notice_fingerprint()
-    |> Client.send_notice()
+    notice =
+      exception
+      |> Notice.new(metadata_with_breadcrumbs, stacktrace, fingerprint)
+      |> put_notice_fingerprint()
+
+    exclude_error_module = Application.get_env(:honeybadger, :exclude_errors)
```
sorentwo, line 296: This seems like a compile-time option rather than a runtime option (not to mention this would be a hot path for an application experiencing a rush of errors). What about pulling this up into a  `compile_env`?

joshuap:
Closing in favor of #342.

PR #443, opened 2022-08-02
Adds config option exclude_errors (#342)

joshuap:
This branch restores #342. See the original PR for summary and initial discussion.

Let's continue discussion on this new PR.


lib/honeybadger.ex
```diff
@@ -287,10 +288,22 @@ defmodule Honeybadger do
...
-    exception
-    |> Notice.new(metadata_with_breadcrumbs, stacktrace, fingerprint)
-    |> put_notice_fingerprint()
-    |> Client.send_notice()
+    notice =
+      exception
+      |> Notice.new(metadata_with_breadcrumbs, stacktrace, fingerprint)
+      |> put_notice_fingerprint()
+
+    exclude_error_value = Application.get_env(:honeybadger, :exclude_errors)
```
joshuap, line 296: Per @sorentwo ([link](https://github.com/honeybadger-io/honeybadger-elixir/pull/442#discussion_r936572264)):

> This seems like a compile-time option rather than a runtime option (not to mention this would be a hot path for an application experiencing a rush of errors). What about pulling this up into a compile_env?

test/honeybadger_test.exs
```diff
@@ -9,6 +9,108 @@ defmodule HoneybadgerTest do
...
+        exception ->
+          :ok = Honeybadger.notify(exception, stacktrace: __STACKTRACE__)
+      end
+
+      assert_receive {:api_request, %{"error" => error}}
+      assert error["class"] == "FunctionClauseError"
+    end
+
+    test "excludes errors sent to server when a list of errors is passed" do
+      restart_with_config(exclude_envs: [], exclude_errors: ["FunctionClauseError"])
```
joshuap, line 99: Is it possible to also support error classes themselves in addition to the string `error.name`? I think some people will configure `exclude_errors: [FunctionClauseError]`.

joshuap (commented):
Looking good @TraceyOnim! A few comments.

joshuap:
This also need a [changelog entry](https://github.com/honeybadger-io/honeybadger-elixir/blob/master/CHANGELOG.md).

test/honeybadger_test.exs
TraceyOnim: Yes, this would be possible . We might need to convert the exception classes to string first then check against the error name. 
```elixir
Enum.map([FunctionClauseError], & &1|> to_string() |> String.trim_leading("Elixir."))
```

lib/honeybadger.ex
TraceyOnim: Fetching the exclude_error using compile_env/2 results to the following error:
```elixir
ERROR! the application :honeybadger has a different value set for key :exclude_errors during runtime compared to compile time. Since this application environment entry was marked as compile time, this difference can lead to different behaviour than expected:

  * Compile time value was not set
  * Runtime value was set to: []

To fix this error, you might:

  * Make the runtime value match the compile time one

  * Recompile your project. If the misconfigured application is a dependency, you may need to run "mix deps.compile honeybadger --force"

  * Alternatively, you can disable this check. If you are using releases, you can set :validate_compile_env to false in your release configuration. If you are using Mix to start your system, you can pass the --no-validate-compile-env flag


** (ErlangError) Erlang error: "aborting bo
```

From the error it suggests we have set a runtime option. Does that mean we also need to set a compile option ? cc @sorentwo

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## [v0.19.0] - 2022-08-10
```
joshuap, line 9: We can add the heading for this when we release (after this PR is merged).

```suggestion
```

joshuap (commented):
LGTM. I realized I can't approve this PR since I'm the author. I want to wait for @sorentwo to review and clarify the compile-time question anyway. 👍

lib/honeybadger.ex
sorentwo: It would have to be referenced in the module body, outside of a function. Don't worry about it, there are several places where we're referencing the application environment during runtime. It's just not a recommended practice.

sorentwo (approved):

joshuap:
@TraceyOnim This just needs documentation in the readme.
cc @subzero10

subzero10:
> @TraceyOnim This just needs documentation in the readme. cc @subzero10

Hey @TraceyOnim, are you working on this?

TraceyOnim:
Yes, I'm working on it

README.md
```diff
@@ -189,6 +189,47 @@ Honeybadger can show arguments in the stacktrace for `FunctionClauseError` excep
 config :honeybadger, filter_args: false
 ```
 
+## Excluding Errors
+
+By default Honeybadger will get notified when error occurs . To override this configuration in order not to send out errors to Honeybadger, set `exclude_errors` option in `config/config.exs`.
```
subzero10, line 194:
```suggestion
By default Honeybadger will be notified when an error occurs. To override this configuration in order not to send out errors to Honeybadger, set `exclude_errors` option in `config/config.exs`.
```

subzero10 (changes requested):

subzero10 (approved):

Issue #408, opened 2021-12-15
The user agent should include the package version


stympy:

joshuap:
Hi @TraceyOnim, the user agent is here:

https://github.com/honeybadger-io/honeybadger-elixir/blob/bff1531643003a03c68b7f8d083e9a1a2f3fe434/lib/honeybadger/client.ex#L11

That string is used as the `User-Agent` header when making requests to our API. What we want is to include the package version info, so our API knows where version of the package is making each request. It should look the same as the `User-Agent` for our Ruby gem, i.e. "HB-Elixir 0.19.0":

https://github.com/honeybadger-io/honeybadger-ruby/blob/f0d3057ad90bcd442ee052aca12811b78c35fbe0/lib/honeybadger/util/http.rb#L21

TraceyOnim:
Thank you @joshuap for the clarification.

PR #455, opened 2022-11-01
Add package version number

TraceyOnim:
related #408

### DIFF
--- a/lib/honeybadger/client.ex
+++ b/lib/honeybadger/client.ex
@@ -52,7 +52,6 @@ defmodule Honeybadger.Client do
       proxy_auth: get_env(opts, :proxy_auth),
       url: get_env(opts, :origin) <> @notices_endpoint
     }
-    |> IO.inspect()
   end
 
   @doc false
--- a/lib/honeybadger/client.ex
+++ b/lib/honeybadger/client.ex
@@ -5,6 +5,11 @@ defmodule Honeybadger.Client do
 
   require Logger
 
+  @headers [
+    {"Accept", "application/json"},
+    {"Content-Type", "application/json"},
+    {"User-Agent", "Honeybadger Elixir #{Honeybadger.Mixfile.project()[:version]}"}
+  ]
   @max_connections 20
   @notices_endpoint "/v1/notices"
 
@@ -47,6 +52,7 @@ defmodule Honeybadger.Client do
       proxy_auth: get_env(opts, :proxy_auth),
       url: get_env(opts, :origin) <> @notices_endpoint
     }
+    |> IO.inspect()
   end
 
   @doc false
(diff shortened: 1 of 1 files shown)


lib/honeybadger/client.ex
```diff
@@ -8,7 +8,7 @@ defmodule Honeybadger.Client do
   @headers [
     {"Accept", "application/json"},
     {"Content-Type", "application/json"},
-    {"User-Agent", "Honeybadger Elixir"}
+    {"User-Agent", "Honeybadger Elixir #{Honeybadger.Mixfile.project()[:version]}"}
```
sorentwo, line 11: This won't change at runtime. Using a module attribute would avoid the extra function call on every invocation.

joshuap (commented):
@TraceyOnim see @sorentwo's review comment.

I'm not sure of how to test this either (I would have to research), but it would be nice if we had a test that sends a notice w/ `Honeybadger.notify` and confirms that the client tries to send the expected request to our servers (the request/response would need to be stubbed). Do you want to look into that?

lib/honeybadger/client.ex
TraceyOnim: Using module attribute that means we will have to edit this module each time the version is changed.

TraceyOnim:
> @TraceyOnim see @sorentwo's review comment.
> 
> I'm not sure of how to test this either (I would have to research), but it would be nice if we had a test that sends a notice w/ `Honeybadger.notify` and confirms that the client tries to send the expected request to our servers (the request/response would need to be stubbed). Do you want to look into that?

I will look into it and let you know the progress

lib/honeybadger/client.ex
sorentwo: Not if you use `@external_attribute` for the mix file. Then it will track changes and force recompilatiin automatically.

TraceyOnim: Hello @sorentwo , I haven't come across `@external_attribute`  before . I was looking into the docs and came across this https://hexdocs.pm/elixir/Module.html#module-external_resource. Is it the same thing?

sorentwo: Yes, that’s the same thing. It forces a recompile when an external file changes.

TraceyOnim: Hello @sorentwo , I have made the change . Let me know if thats how it should be

sorentwo (commented):
This looks perfect. My only request is a test that verifies the version is present. If there was a typo in here, say it used `[:verson]`, this would still "work" and print "Honeybadger Elixir " without any kind of error.

TraceyOnim:
> This looks perfect. My only request is a test that verifies the version is present. If there was a typo in here, say it used `[:verson]`, this would still "work" and print "Honeybadger Elixir " without any kind of error.

Okay, working on that

subzero10 (approved):

Issue #467, opened 2023-02-06
HTTP client options


danschultzer:
I've experienced issues with honeybadger not working in a IPv6 only network. This is due to a long running issue in hackney of how IP addresses are resolved: https://github.com/benoitc/hackney/issues/206 (in [particular this logic](https://github.com/benoitc/hackney/blob/6e79b2bb11a77389d3ba9ff3a0828a45796fe7a8/src/hackney_util.erl#L81-L91)).

Looking at how honeybadger deals with HTTP it would be good to:

1. Allow options to be passed to hackney (~~this won't solve my particular issue though~~ edit: should work)
2. Allow option to chose HTTP client
  a. Switch to a flexible library like Tesla
  b. Add an abstraction in honeybadger (this is how I handle it in https://github.com/pow-auth/assent/tree/main/lib/assent to support built-in httpc by default)

A third option, which is about solving the IPv6 issue, is to switch to something other than hackney since upstream fix seems unlikely to happen anytime soon. I solved my issue by switching to Finch and pass in `conn_opts`: https://github.com/danschultzer/honeybadger-elixir/commit/14bdbe21f6ea754c335478ccb82bb6fa312993a7

danschultzer:
Sounds great @sorentwo, thanks for looking at this!

I did make a quick PR in #468 to pass hackney opts which solves part of the issue. But it would be nice to just pass in an adapter to deal with edge cases and have more control over the connection egress.

Just a warning on `:httpc` that the cert check has to be handled explicitly (not built-in 😞). But apart from that and no HTTP/2 support, it works fine, and the benefit is that it helps limit the dependency graph size.

fastjames:
Is there still any interest in implementing this change? As of today hackney has a low-priority CVE with no fix listed (and no obvious fix activity on github). I'd love to swap it out for a different client if possible.

fastjames:
> [@fastjames](https://github.com/fastjames) Yup!

Well, it's good to know that someone else is interested as well. I read over some of the earlier links on this issue, and @danschultzer 's work in assent looked like it might work here too. I took a stab at shamelessly copying what he did, and I think I'm getting somewhere (see https://github.com/fastjames/honeybadger-elixir/tree/fastjames/feature/multi_http_client_support for my work so far).

Important questions still to be answered:

1. Is hackney support still important? I'm working on that now, but if literally no one wants it then 💥 .
[...]

PR #468, opened 2023-02-06
Allow for hackney options

danschultzer:
This PR makes it possible to set `:hackney_opts` in the configuration which is passed onto hackney. I didn't see any tests of configuration, but this works in local for me. This resolves the first part of #467.

To force IPv6 with hackney I can now set the config with:

```elixir
config :honeybadger,
  # ...
  hackney_opts: [connect_options: [:inet6]]
```

But dual stack could still be an issue since it'll always default to IPv4 in hackney.

As discussed in #467 it would be better if it was possible to pass in a HTTP client adapter. That would remove things that arguably should be outside the scope of this library, like the HTTP proxy tunnel or the connection pool. The default startup could still include hackney with some sane defaults (e.g. starting connection pool).

### DIFF
--- a/lib/honeybadger/client.ex
+++ b/lib/honeybadger/client.ex
@@ -21,7 +21,8 @@ defmodule Honeybadger.Client do
           headers: [{binary(), term()}],
           proxy: binary(),
           proxy_auth: {binary(), binary()},
-          url: binary()
+          url: binary(),
+          hackney_opts: keyword()
         }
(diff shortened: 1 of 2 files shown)


sorentwo (approved):
These changes look great as a stop-gap until we have pluggable clients. I'm going to merge this in for now, but the changes may be overwritten with client changes before the next release.

Issue #554, opened 2024-06-19
Support Reporting Events for Honeybadger Insights


mdg:
[Honeybadger Insights](https://www.honeybadger.io/tour/logging-observability/) looks really cool. It seems to allow searching items uploaded from notices, but also supports events. The elixir SDK appears to only support notices right now. Would be great if it also supports reporting events.
https://docs.honeybadger.io/api/reporting-events

stympy:
Thanks for the issue! We definitely want to add support for events, but we don't have an ETA at the moment. PRs are welcome, of course. 😉 

This is the implementation in the Ruby gem, and we'd want to follow a similar pattern here:

```ruby
Honeybadger.event("event name", example_payload: "something something")
Honeybadger.event(event_type: "event name", example_payload: "something something")

# signature:
Honeybadger.event(event_type, payload={})
Honeybadger.event(payload={})
```

The event payloads should look like this:

```json
{
	"ts": "2023-12-07T19:53:12.502",
	"event_type": "created.user",
	"email": "bob@example.com"
}
```

mdg:
I posted a PR to add this capability. I've deployed it to a low traffic site some events have been uploaded. Happy to share the project ID if that's helpful.

mdg:
Thanks for your help with this!

PR #564, opened 2024-08-28
add Honeybadger.event interface

mdg:
send application events to the event endpoint
add a timestamp

issue #554
https://github.com/honeybadger-io/honeybadger-elixir/issues/554

### DIFF
--- a/lib/honeybadger.ex
+++ b/lib/honeybadger.ex
@@ -344,7 +344,7 @@ defmodule Honeybadger do
     ts = DateTime.utc_now() |> DateTime.to_string()
 
     event_data
-    |> Map.put(:ts, ts)
+    |> Map.put_new(:ts, ts)
     |> Client.send_event()
   end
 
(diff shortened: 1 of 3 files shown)


lib/honeybadger.ex
```diff
@@ -332,6 +332,20 @@ defmodule Honeybadger do
...
+    ts = DateTime.utc_now() |> DateTime.to_string()
+    event_data
+    |> Map.put(:event_type, event_type)
+    |> Map.put(:ts, ts)
+    |> Client.send_event()
+  end
+
+  @spec event(map()) :: :ok
+  def event(event_data) do
+    Client.send_event(event_data)
```
stympy, line 346: Could you supply a `ts` of `DateTime.utc_now()` here if one isn't present in the map?

stympy (approved):
This is looking good, thanks! Could you address the `ts` comment and run `mix format`?

mdg:
@stympy thanks for the helpful prompt review. I think those 2 issues should be addressed now.

stympy (changes requested):
Looking good. I have some [suggested changes](https://github.com/pathpub/honeybadger-elixir/pull/1) for you. :)

stympy (approved):

stympy:
Released in 0.22.0. Thanks!

mdg:
I've upgraded our app back to the main version. Thanks for handling it so quickly!

PR #579, opened 2025-02-13
Add support for multiple HTTP client libraries

### LINKED ISSUE #467: HTTP client options

fastjames:
In order to avoid library lock-in and allow customers to use the HTTP client library that best fits their needs, introduce the `HTTPAdapter` pattern with initial support for [req](https://github.com/wojtekmach/req) (default), [finch](https://github.com/sneako/finch), and [:hackney](https://github.com/benoitc/hackney).

closes #467


fastjames:
The most recent test failures happen for me locally with following elixir/OTP version combos:
* 1.15.8-otp-26 / 26.2.5.8
* 1.15.8-otp-25 / 25.3.2.17

I also tried 1.15.8-otp-24 / 24.3.4.17 but this OTP appears to be missing some required telemetry-related capabilities so it wouldn't run at all.

I found a few references to the errors seen here in issues on erlang/otp:
* https://github.com/erlang/otp/issues/7493
* https://github.com/erlang/otp/issues/7774

I read through the docs for [test_server](https://github.com/danschultzer/test_server) and [cowboy](https://github.com/ninenines/cowboy) but I haven't yet found a way to configure which TLS versions the test server supports so that I can test the hypotheses contained in the linked erlang/otp issues.

sorentwo:
@fastjames Will you re-push to trigger a new build? The logs are gone and I can't see what the original errors were. I'm confident there is some configuration or package changes we can make to ensure Elixir v1.15 support.

fastjames:
> @fastjames Will you re-push to trigger a new build? The logs are gone and I can't see what the original errors were. I'm confident there is some configuration or package changes we can make to ensure Elixir v1.15 support.

I have reordered the last 2 commits and pushed, so I think that did the trick.

fastjames:
@sorentwo Here is the downloaded log archive from the failing 1.15 test. Hopefully this contains the information we need to fix it up.
[logs_34502181737.zip](https://github.com/user-attachments/files/18848470/logs_34502181737.zip)

lib/honeybadger/http_adapter.ex
```diff
@@ -0,0 +1,173 @@
...
+      [request_url | _rest] = String.split(response.request_url, "?", parts: 2)
+
+      """
+      HTTP Adapter: #{inspect(response.http_adapter)}
+      Request URL: #{request_url}
+
+      Response status: #{response.status}
+
+      Response headers:
+      #{Enum.reduce(response.headers, "", fn {k, v}, acc -> acc <> "\n#{k}: #{v}" end)}
```
sorentwo, line 57:
```suggestion
      #{Enum.map_join(response.headers, "\n", fn {key, val} -> "#{key}: #{val}" end)}
```

lib/honeybadger/http_adapter.ex
```diff
@@ -0,0 +1,173 @@
...
+
+  @doc """
+  Sets a user agent header.
+
+  The header value will be `Honeybadger-VERSION` with VERSION being the `:vsn` of
+  `:honeybadger` app.
+  """
+  @spec user_agent_header() :: {binary(), binary()}
+  def user_agent_header do
+    version = Application.spec(:honeybadger, :vsn) || "0.0.0"
```
sorentwo, line 80: The application has to be loaded before this call is made, so I think it can safely omit the `||`:

```suggestion
    version = Application.spec(:honeybadger, :vsn)
```

mix.exs
```diff
@@ -83,7 +83,9 @@ defmodule Honeybadger.Mixfile do
 
   defp deps do
     [
-      {:hackney, "~> 1.1"},
+      {:finch, "~> 0.19", optional: true},
```
sorentwo, line 86: Since we want to nudge people toward `req`, and I suspect _very_ few applications would use `finch` directly without `req`, I'd vote to omit a dedicated `finch` adapter entirely.

lib/honeybadger/http_adapter/req.ex
```diff
@@ -0,0 +1,49 @@
...
+    See `Honeybadger.HTTPAdapter` for more.
+    """
+    alias Honeybadger.{HTTPAdapter, HTTPAdapter.HTTPResponse}
+
+    @behaviour HTTPAdapter
+
+    @impl HTTPAdapter
+    def request(method, url, body, headers, req_opts \\ nil) do
+      headers = headers ++ [HTTPAdapter.user_agent_header()]
+      opts = req_opts || []
```
sorentwo, line 19: Can we make an empty list the default?

```suggestion
    def request(method, url, body, headers, req_opts \\ []) do
      headers = headers ++ [HTTPAdapter.user_agent_header()]
```

lib/honeybadger/http_adapter/req.ex
```diff
@@ -0,0 +1,49 @@
...
+      |> Req.new()
+      |> Req.request()
+      |> case do
+        {:ok, response} ->
+          headers =
+            Enum.map(headers, fn {key, value} ->
+              {String.downcase(to_string(key)), to_string(value)}
+            end)
+
+          {:ok, %HTTPResponse{status: response.status, headers: headers, body: response.body}}
```
sorentwo, line 42: Req [stores the header names downcased](https://hexdocs.pm/req/Req.html#module-header-names) already, so this shouldn't be necessary.

```suggestion
          {:ok, %HTTPResponse{status: response.status, headers: response.headers, body: response.body}}
```

lib/honeybadger/http_adapter.ex
```diff
@@ -0,0 +1,173 @@
...
+    end
+    |> case do
+      {:ok, %{status: status} = resp} when status in 200..399 ->
+        {:ok, %{resp | http_adapter: http_adapter, request_url: url}}
+
+      {:ok, %{status: status} = resp} when status in 400..599 ->
+        {:error, %{resp | http_adapter: http_adapter, request_url: url}}
+
+      {:error, error} ->
+        {:error, error}
```
sorentwo, line 131: WDYT about inlining these into the case above, rather than piping into two cases?

lib/honeybadger/http_adapter.ex
```diff
@@ -0,0 +1,173 @@
...
+  @doc """
+  Decodes request response body.
+  """
+  @spec decode_response(HTTPResponse.t(), Keyword.t()) ::
+          {:ok, HTTPResponse.t()} | {:error, InvalidResponseError.t()}
+  def decode_response(%HTTPResponse{} = response, opts) do
+    case decode(response.headers, response.body, opts) do
+      {:ok, body} -> {:ok, %{response | body: body}}
+      {:error, _error} -> {:error, InvalidResponseError.exception(response: response)}
+    end
```
sorentwo, line 153: Req already decodes responses automatically. This could be inlined into `hackney` if `finch` is removed.

lib/honeybadger/http_adapter.ex
```diff
@@ -0,0 +1,173 @@
...
+
+      {:error, error} ->
+        {:error, error}
+    end
+  end
+
+  defp get_adapter(opts) do
+    default_http_adapter = Application.get_env(:honeybadger, :http_adapter, @default_http_client)
+
+    case Keyword.get(opts, :http_adapter, default_http_adapter) do
```
sorentwo, line 138: The difference between `default_http_client` and `default_http_adapter` tripped me up briefly.

test/honeybadger/http_adapter/req_test.exs
```diff
@@ -0,0 +1,114 @@
...
+        Keyword.put(
+          @req_opts,
+          :connect_options,
+          transport_opts: [cacerts: TestServer.x509_suite().cacerts],
+          protocols: [:http2]
+        )
+
+      assert {:ok, %HTTPResponse{status: 200, body: "HTTP/2"}} =
+               Req.request(:get, TestServer.url(), nil, [], req_opts)
+    end
```
sorentwo, line 26: This is testing the functionality of `req`, and thereby `finch` and `mint`, which seems outside the scope of what the honeybadger library should care about.

sorentwo (commented):
This is an impressive effort! I've left a few small suggestions, but I also think a little could be removed. Namely, the stand-alone `finch` adapter.

mix.exs
fastjames: This is a good observation, I'll remove that adapter.

lib/honeybadger/http_adapter.ex
fastjames: Would @default_http_client_lib be a clearer name?

lib/honeybadger/http_adapter.ex
fastjames: This makes a lot of sense, so I have moved the decoding code into the hackney adapter module. 👍

lib/honeybadger/http_adapter.ex
fastjames: I took a shot at this and collapsing them into one `case` didn't feel right. It seemed to work out better as a `with`. That required me to flip the error handling around, so I might need to make more changes / cover more cases there.

test/honeybadger/http_adapter/req_test.exs
fastjames: OK, I can see removing this test for that reason. I don't mind it being there to keep parity with other adapter test suites (e.g. hackney) but you're right that this one isn't testing much code in the library itself.

stympy (approved):
It would be nice to nail down the test failures, but otherwise I'm good with this :)

stympy:
Assigned to @rabidpraxis to merge when ready :)

rabidpraxis:
Hey @fastjames! Thanks again for adding HTTP adapters, it is an awesome addition!

I pulled this down, played around with it, and made a few updates:

* I removed the SSL tests for both adapters, as they were not really testing any of the adapter functionality but more the compatibility of TestServer and the underlying request library.
* I ran into issues with hackney compile time warnings (if I didn't have it as a dependency), so I wrapped the module in the same `ensure_loaded?` check as `Req`. I then noticed that if you add an HTTP client library dep after installing Honeybadger, Honeybadger would not recompile, and the module would still not be loaded. As a way to deal with that, I refactored things so we load and check for the library at runtime, but we check for it's existence during `Honeybadger.start/2`, so we throw before the app fully boots. I memoized the default client lookup with `:persistent_term` to keep from calling `code_loaded?` too many times.
* I added a specific message if we found no configured/loaded HTTP clients vs if the client was configured but missing its dependency
* I noticed that the "User-Agent" was being set twice: once in the `Client` and then in the adapter, so I removed the header update in the adapter.
* Fixed a small issue where Req would fail if no opts were passed in the Honeybadger config override
* Added some details to the README

PR #580, opened 2025-02-14
Update test matrix

stympy:
Drop elixir 1.14 and 1.15. Add elixir 1.16 and 1.18.

### DIFF
--- a/.github/workflows/elixir.yml
+++ b/.github/workflows/elixir.yml
@@ -18,13 +18,13 @@ jobs:
       matrix:
         include:
           - pair:
-              elixir: '1.14'
-              otp: '23.3'
-          - pair:
-              elixir: '1.15'
+              elixir: '1.16'
               otp: '26.0'
           - pair:
               elixir: '1.17'
+              otp: '27.0'
+          - pair:
+              elixir: '1.18'
               otp: '27.0'
             lint: lint
 
--- a/.github/workflows/elixir.yml
+++ b/.github/workflows/elixir.yml
@@ -31,7 +31,7 @@ jobs:
               otp: '27.0'
             lint: lint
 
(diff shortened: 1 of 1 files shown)


stympy:
#579 is having problems with Elixir 1.15, and https://endoflife.date/elixir says that support for 1.15 ended more than a year ago, so I think it's safe to bump up our minimum requirements.

joshuap:
> #579 is having problems with Elixir 1.15, and https://endoflife.date/elixir says that support for 1.15 ended more than a year ago, so I think it's safe to bump up our minimum requirements.

Our official policy is to support all versions that still receive security patches (e.g. until they reach EOL).

https://hexdocs.pm/elixir/1.18.2/compatibility-and-deprecations.html

joshuap:
That is, I'm not necessarily against removing 1.14/1.15 from the test matrix if they are causing problems, but we do still need to support them until they reach EOL.

stympy:
We could release 0.23 with this change, then branch off 0.22.1 if we need to do update that supports the older elixir versions.

.github/workflows/elixir.yml
```diff
@@ -18,14 +18,14 @@ jobs:
       matrix:
         include:
           - pair:
-              elixir: '1.14'
-              otp: '23.3'
-          - pair:
-              elixir: '1.15'
+              elixir: '1.16'
```
sorentwo, line 21: Elixir 1.18 came out rather recently. New versions are released roughly every 6 months, and v1.15 is only a year and a half old. IMO the minimum supported version should be at least Elixir v1.15 and OTP 24.3.

sorentwo (approved):
One other small thing, you should probably update `ubuntu` as well. That version is EOL and won't work in a few months.

PR #591, opened 2025-02-28
Insights instrumentation

rabidpraxis:
This still needs documentation, but I wanted to get some feedback on the general direction. I added a set of modules focused on automatic instrumentation which shares some properties with our Ruby plugins. Here is the Oban one:

https://github.com/honeybadger-io/honeybadger-elixir/blob/3cfda1707553faf86f6ce3812c1fec446be07a2e/lib/honeybadger/insights/oban.ex#L1-L28

This is disabled by default via the `insights_enabled` config option.

If enabled, all modules are attempted to be attached, but will skip loading if any of the `@requred_dependencies` return false from `Code.ensure_loaded?/1`. Each module specifies which events they want to subscribe to via `@telemetry_events` list and then an `extract_metadata/2` callback must be implemented to, well, extract metadata for the event.

There are currently 2 ways to customize the instrumented events. You can add your own `event_filter` by implementing the `Honeybadger.EventFilter.filter/3` callback and overriding it in the config. All instrumented events will funnel through this. The telemetry event name is passed as the third argument so users can just filter on a specific event type:

```elixir
defmodule MyEventFilter do
  def filter(event, raw, "oban.job.stop") do
    Map.put(event, :thing, raw.thing)
  end

  def filter(event, _, _), do: event
end
```

I also added a way to customize each of these "plugins". There is a `insights_config` option that accepts a map. Users can set the telemetry events per "plugin". Say you only wanted one live_view event to be logged to Insights. Instead of adding a filter module, you could:

```elixir
config :honeybadger,
  insights_config: %{
    live_view: %{
      telemetry_events: [
        "phoenix.live_component.handle_event.stop"
      ]
    }
  }
```

Some of the plugins include their own options. For example, this can be used to customize the excluded queries from Ecto:

```elixir
config :honeybadger,
  insights_config: %{
    ecto: %{
      excluded_queries: [
        ~r/SELECT 1/
      ]
    }
  }
```

There is no spec around any of these options, so it's pretty loose. I chose to go this way instead of adding each possible option. I wasn't sure how idiomatic of a pattern this is, so any feedback about this is welcome.


sorentwo:
@rabidpraxis A little swamped this past week, but I can review this by tomorrow.

lib/honeybadger/insights/ecto.ex
```diff
@@ -0,0 +1,56 @@
...
+    "oban_peers"
+  ]
+
+  def get_telemetry_events do
+    Honeybadger.get_env(:ecto_repos)
+    |> Enum.map(&get_telemetry_prefix/1)
+  end
+
+  def extract_metadata(meta, _name) do
+    Map.take(meta, [:query, :decode_time, :query_time, :queue_time, :source])
```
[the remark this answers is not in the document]
rabidpraxis: Yeah, we probably should... I'll add that.

lib/honeybadger/insights/absinthe.ex
```diff
@@ -0,0 +1,99 @@
...
+
+  defp get_operation_type(_), do: nil
+
+  defp get_operation_name(%{blueprint: blueprint}) when is_map(blueprint) do
+    operation = current_operation(blueprint)
+
+    case operation do
+      nil -> nil
+      operation -> Map.get(operation, :name)
+    end
```
sorentwo, line 84: This pattern is used a few times. Could be simpler to use `get_in/2`, as long as it's not using structs because it's `nil` safe:

```suggestion
    blueprint |> current_operation() |> get_in([:name])
```

lib/honeybadger/insights/base.ex
```diff
@@ -0,0 +1,175 @@
...
+      def dependencies_available? do
+        # We are using this strategy to deal with a compiler warning about the
+        # empty case never being reached.
+        deps = @required_dependencies
+
+        if Enum.empty?(deps) do
+          true
+        else
+          Enum.all?(deps, fn mod -> Code.ensure_loaded?(mod) end)
+        end
```
sorentwo, line 44: `Enum.all?` returns true for an empty list:

```suggestion
        Enum.all?(deps, fn mod -> Code.ensure_loaded?(mod) end)
```

lib/honeybadger/insights/base.ex
```diff
@@ -0,0 +1,175 @@
...
+  defmacro __before_compile__(env) do
+    namespace =
+      env.module
+      |> Atom.to_string()
+      |> String.split(".")
+      |> List.last()
+      |> Macro.underscore()
+      |> String.to_atom()
+
+    quote do
```
sorentwo, line 27: This generates a lot of content for each module that's paramaterizable. Usually, the preferred approach is to generate wrappers that call functions on the parent module. For example,

```elixir
def dependencies_available? do
  Base.dependencies_available?(@required_dependencies)
end
```

lib/honeybadger/insights/base.ex
```diff
@@ -0,0 +1,175 @@
...
+                   :query_time,
+                   :queue_time,
+                   :idle_time
+                 ] ->
+              Map.put(acc, key, System.convert_time_unit(value, :native, :millisecond))
+
+            _ ->
+              Map.put(acc, key, value)
+          end
+        end)
```
sorentwo, line 130: Can be done with a single reduce to avoid iterating through the measurements twice (but I extracted `@time_keys` into a module attribute)

```suggestion
        Enum.reduce(measurements, %{}, fn
          {key, _vl}, acc when key in ~w(monotonic_time end_time_mono)a ->
            acc

          {key, val}, acc when key in @time_keys ->
            Map.put(acc, key, System.convert_time_unit(val, :native, :millisecond))

          {key, val}, acc ->
            Map.put(acc, key, val)
        end)
```

lib/honeybadger/insights/ecto.ex
```diff
@@ -0,0 +1,55 @@
...
+
+  @excluded_sources [
+    "schema_migrations",
+    "oban_jobs",
+    "oban_peers"
+  ]
+
+  def get_telemetry_events do
+    Honeybadger.get_env(:ecto_repos)
+    |> Enum.map(&get_telemetry_prefix/1)
```
sorentwo, line 24:
```suggestion
    :ecto_repos
    |> Honeybadger.get_env()
    |> Enum.map(&get_telemetry_prefix/1)
```

lib/honeybadger/insights/ecto.ex
```diff
@@ -0,0 +1,55 @@
...
+      true
+    else
+      get_insights_config(:excluded_queries, @excluded_queries)
+      |> Enum.any?(fn pattern ->
+        case pattern do
+          pattern when is_binary(pattern) -> query == pattern
+          %Regex{} -> Regex.match?(pattern, query)
+          _ -> false
+        end
+      end)
```
sorentwo, line 42: Nitpick only (avoiding single pipes and nested case):

```suggestion
      :excluded_queries
      |> get_insights_config(@excluded_queries)
      |> Enum.any?(fn
          pattern when is_binary(pattern) -> query == pattern
          %Regex{} = pattern -> Regex.match?(pattern, query)
          _pattern -> false
        end)
```

lib/honeybadger/insights/ecto.ex
```diff
@@ -0,0 +1,55 @@
...
+
+  defp get_telemetry_prefix(repo) do
+    case Keyword.get(repo.config(), :telemetry_prefix) do
+      nil ->
+        []
+
+      telemetry_prefix ->
+        telemetry_prefix ++ [:query]
+    end
+  end
```
sorentwo, line 54: Keyword.get takes a default:

```suggestion
    Keyword.get(repo.config(), :telemetry_prefix, []) ++ [:query]
```

lib/honeybadger/insights/oban.ex
```diff
@@ -0,0 +1,28 @@
...
+      :memory,
+      :prefix,
+      :queue,
+      :queue_time,
+      :state,
+      :tags,
+      :time,
+      :worker
+    ])
+  end
```
sorentwo, line 27: All fields that are defined on a job are undocumented and deprecated. They should be taken off the job instead. Some of these attributes don't exist, and others are actually from `measurements`:

```suggestion
  def extract_metadata(%{conf: conf, job: job, state: state}, _name) do
    job
    |> Map.take(~w(args attempt id queue tags worker)a)
    |> Map.put(:prefix, conf.prefix)
    |> Map.put(:state, state)
  end
```

lib/honeybadger/utils.ex
```diff
@@ -18,6 +18,20 @@ defmodule Honeybadger.Utils do
...
+
+  # Example
+
+      iex> Honeybadger.Utils.dotify([:Honeybadger, :Utils])
+      "Honeybadger.Utils"
+  """
+  def dotify(path) when is_list(path) do
+    path
+    |> Enum.map(&to_string/1)
+    |> Enum.join(".")
```
sorentwo, line 32:
```suggestion
    Enum.map_join(path, ".", &to_string/1)
```

sorentwo (commented):
Great addition to HB, and I'm keen on the chosen list of packages 😉.

lib/honeybadger/insights/tesla.ex
```diff
@@ -0,0 +1,22 @@
...
+    "tesla.request.stop",
+    "tesla.request.exception"
+  ]
+
+  def extract_metadata(meta, _name) do
+    env = Map.get(meta, :env, %{})
+
+    %{
+      method: env.method |> to_string() |> String.upcase(),
+      url: env.url,
```
roelbondoc, line 18: In the [ruby implementation](https://github.com/honeybadger-io/honeybadger-ruby/blob/master/lib/honeybadger/plugins/net_http.rb#L46), we default to recording the hostname only. We enable the full url by setting `net_http.insights.full_url` to true. Is it possible to do that here?

rabidpraxis: Yeah, we can add that.

lib/honeybadger/insights/oban.ex
rabidpraxis: 👍🏻 I think I pulled these out while inspecting the instrumentation payload, so I was unaware of what is deprecated.

@sorentwo You are emitting a gaggle of telemetry events in Oban. Are there any others that would be useful to include by default?

lib/honeybadger/insights/base.ex
rabidpraxis: Is this more what you were thinking about?

https://github.com/honeybadger-io/honeybadger-elixir/pull/591/commits/0ad97009fda823273a0d15acb01d9d44951470eb

lib/honeybadger/insights/base.ex
rabidpraxis: Should be addressed in https://github.com/honeybadger-io/honeybadger-elixir/commit/0ad97009fda823273a0d15acb01d9d44951470eb

lib/honeybadger/insights/ecto.ex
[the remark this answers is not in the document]
rabidpraxis: https://github.com/honeybadger-io/honeybadger-elixir/pull/591/commits/54504870f1e27d20cd5eb1acc02054e1461b3e65

lib/honeybadger/insights/absinthe.ex
rabidpraxis: Yeah, I keep getting the `Access behaviour` error when I use get_in with most of the metadata coming back from Absinthe.

lib/honeybadger/insights/ecto.ex
```diff
@@ -0,0 +1,75 @@
...
+
+  defp get_telemetry_prefix(repo) do
+    Keyword.get(repo.config(), :telemetry_prefix, []) ++ [:query]
+  end
+
+  @escape_quotes ~r/(\\\"|\\')/
+  @squote_data ~r/'(?:[^']|'')*'/
+  @dquote_data ~r/"(?:[^"]|"")*"/
+  @number_data ~r/\b\d+\b/
+  @double_quoters ~r/(postgres|sqlite3|postgis)/i
```
rabidpraxis, line 56: I think it's more idiomatic to keep these kinds of variables at the top of the module, but they are only used in the function below so I thought it made sense to keep them close.

lib/honeybadger/insights/live_view.ex
```diff
@@ -0,0 +1,44 @@
...
+  defp extract_view(%{socket: socket}) do
+    socket.view |> get_module_name()
+  rescue
+    _ -> nil
+  end
+
+  defp extract_view(_), do: nil
+
+  defp extract_assigns(%{socket: socket}) do
+    socket.assigns
```
roelbondoc, line 34: Not sure if this is a concern or not, feel free to ignore.

When this data gets injected into the event, it's possible that it could include too much or unwanted data. Should we make this optional? Or is there a way to represent the data in another form?

rabidpraxis: The data also gets run through the same filters as the params (I think of assigns as the params of liveview events), which is also why I updated the filtered params to remove some things from the assigns:

https://github.com/honeybadger-io/honeybadger-elixir/blob/4dc8bef4fc2aaae285d079989c6c2e1f76767900/mix.exs#L74

roelbondoc (approved):
Sorry I don't have much to add in terms of code changes and implementation. But from what I've tested this works well! Nice work!

lib/honeybadger/insights/tesla.ex
rabidpraxis: https://github.com/honeybadger-io/honeybadger-elixir/pull/591/commits/d83362d2609e4abab35f57aa8946b1de022a3852

rabidpraxis:
Okay, I added some basic tests for each module and documentation (both at the module level and in the Readme). I think this is ready to go if there are no other comments.

README.md
```diff
@@ -150,6 +150,117 @@ rescue
...
+  insights_enabled: true
+```
+
+### Supported Libraries
+
+Honeybadger automatically instruments the following libraries when they are available:
+
+* Absinthe: GraphQL query execution
+* Ecto: Database queries
+* Finch: HTTP client requests
```
sorentwo, line 170: Few apps use Finch directly, it's usually through a wrapper like Req. That may be helpful to mention here when people are scanning for supported packages.

lib/honeybadger/client.ex
```diff
@@ -79,6 +79,7 @@ defmodule Honeybadger.Client do
   Upload the event data
   """
   @spec send_event(map) :: :ok
+
```
sorentwo, line 82: There shouldn't be space between the spec and def here

```suggestion
```

lib/honeybadger/insights/base.ex
```diff
@@ -0,0 +1,164 @@
...
+  Base module providing common telemetry attachment functionality.
+  """
+
+  defmacro __using__(_opts) do
+    quote do
+      require Logger
+
+      # Define empty defaults for the attributes
+      Module.register_attribute(__MODULE__, :required_dependencies, [])
+      @required_dependencies []
```
sorentwo, line 12: This isn't needed, the register already initialized it.

```suggestion
```

lib/honeybadger/insights/base.ex
```diff
@@ -0,0 +1,164 @@
...
+        get_insights_config(:telemetry_events, @telemetry_events)
+      end
+
+      @doc """
+      Attaches telemetry handlers if all required dependencies are available.
+      """
+      def attach do
+        if dependencies_available?() do
+          get_telemetry_events()
+          |> Enum.each(&attach_event/1)
```
sorentwo, line 72: Same "single pipe" nitpick:

```suggestion
          Enum.each(get_telemetry_events(), &attach_event/1)
```

lib/honeybadger/insights/base.ex
```diff
@@ -0,0 +1,164 @@
...
+          {:error, :missing_dependencies}
+        end
+      end
+
+      @doc """
+      Attaches a telemetry handler for a specific event.
+      The event can be either a string (dot-separated) or a list of atoms.
+      """
+      def attach_event(event) when is_binary(event) do
+        event_atoms = event |> String.split(".") |> Enum.map(&String.to_atom/1)
```
sorentwo, line 89: Use of `String.to_atom/1` isn't considered safe and this will get flagged by security scanners. Use `to_existing_atom/1` instead

```suggestion
        event_atoms = event |> String.split(".") |> Enum.map(&String.to_existing_atom/1)
```

lib/honeybadger/insights/base.ex
```diff
@@ -0,0 +1,164 @@
...
+      """
+      def handle_telemetry(event_name, measurements, metadata, _opts) do
+        name = Honeybadger.Utils.dotify(event_name)
+
+        unless ignore?(metadata) do
+          %{event_type: name}
+          |> Map.merge(process_measurements(measurements))
+          |> Map.merge(
+            extract_metadata(metadata, name)
+            |> Map.reject(fn {_, v} -> is_nil(v) end)
```
sorentwo, line 132: Single pipe:

```suggestion
            metadata
            |> extract_metadata(name)
            |> Map.reject(fn {_, v} -> is_nil(v) end)
```

lib/honeybadger/insights/finch.ex
```diff
@@ -0,0 +1,78 @@
...
+      {:error, error} ->
+        Map.put(metadata, :error, Exception.message(error))
+    end
+  end
+
+  defp reconstruct_url(request) do
+    # Exclude query parameters for security reasons
+    port_string = get_port_string(request.scheme, request.port)
+
+    "#{request.scheme}://#{request.host}#{port_string}#{request.path}"
```
sorentwo, line 68: It may be a bit safer to use the URI module here, but I don't think it's critical

test/honeybadger/insights/absinthe_test.exs
```diff
@@ -0,0 +1,156 @@
+defmodule Honeybadger.Insights.AbsintheTest do
+  use Honeybadger.Case, async: false
```
sorentwo, line 2: Is there something in here that prevents this from being async? Seems rather isolated because there isn't any config changes (that I see)

test/test_helper.exs
```diff
@@ -8,12 +8,15 @@ Application.put_all_env(
...
     api_key: "abc123",
-    origin: "http://localhost:4444"
+    origin: "http://localhost:4444",
+    insights_enabled: true
   ]
 )
 
 ExUnit.start(assert_receive_timeout: 1000, refute_receive_timeout: 1000)
 
+Code.require_file("test/support/insights_case.ex")
```
sorentwo, line 18: Usually, the support directory is configured to be included automatically and you don't have a manual require.

sorentwo (approved):
Very impressive work here! I've left some nitpicks, but nothing blocking aside from the to_atom usage.

lib/honeybadger/insights/base.ex
rabidpraxis: Things seem to break when I remove that initial declaration. Even though I have the comment (which might have been AI) I don't think the third argument of `register_attribute/3` is [the default value](https://hexdocs.pm/elixir/Module.html#register_attribute/3), which makes me wonder if I even need to declare it.

test/test_helper.exs
rabidpraxis: Ah yeah, [this](https://github.com/honeybadger-io/honeybadger-elixir/pull/591/commits/def0c08fd8bcd31bba1dbc45d7398952de585e3f) is better.

README.md
rabidpraxis: Thanks for making me aware of that. I updated the comment. I also looked into who else uses Finch, and it looks like Tesla has an adapter. I added some code to not send the Tesla insights event if [Finch is used as the adapter](https://github.com/honeybadger-io/honeybadger-elixir/pull/591/commits/0dfe1f27605eceba12f83e878a6d4690ad51bd73).

test/honeybadger/insights/absinthe_test.exs
rabidpraxis: Nope, oversight.

lib/honeybadger/insights/base.ex
rabidpraxis: I made this change but I had to add some atoms to the start of the test_helper, as they were not available. 

@sorentwo This makes me wonder if using `to_existing_atom` could raise errors for some users, perhaps if they have an early version of a library that does not include the atoms used in the telemetry subscription? What if we just ignore the credo check with a prepending comment, as the strings should only be coming from the module var or the application config.

README.md
sorentwo: It may cause errors depending on the load order, especially for dynamically loaded modules in development. Declaring the list of atoms is a safe work around.

test/honeybadger/insights/absinthe_test.exs
rabidpraxis: I realized you were talking about more than just Absinthe so I enabled async for all the modules and things started failing. Since we are starting and stopping the test_helper running these tests in parallel throws an "application already running" error.

PR #597, opened 2025-03-20
Add Request ID to Insights events and notices

rabidpraxis:
This is the last piece of the Insights instrumentation work. I based this on the instrumentation branch, as I used telemetry for a good bit of the implementation.

We don't support full blown traces in Insights yet but we do want to allow users to correlate their structured logs. We do that by propagating a `request_id` into all events, if available. 

Here I focused on:

- Plug and Phoenix HTTP request processes
- LiveView(Component) events
  - I decided to generate a new request_id for live_view & live_component events, as I think of those like the "request" of a LiveView. LiveView events are spawned in a shared mounted process, so I had to clear the request_id at the end of the event (which I don't have to worry about with Plug requests)
- Oban jobs
  - @sorentwo, I tried a few ways to make this easy on the user (like adding a module to be used in the worker that "overrode" `new`) but all the methods I attempted had some tradeoff that made me feel like it was the wrong direction. I landed on a simple function that can be used to [augment the Changeset](https://github.com/honeybadger-io/honeybadger-elixir/blob/a159c3c2ef1afde0b3414424cedeeda05a5dfa6b/lib/honeybadger/insights/oban.ex#L38-L51) before it is inserted. Curious to hear your thoughts.

I also looked for simple ways to propagate the `request_id` to things like GenServers or Tasks, but at this point I think it might be simplest to document a process for passing a request_id (say into a GenServer cast) to ensure it gets into the process dictionary with our helpers.


lib/honeybadger/utils.ex
```diff
@@ -45,6 +45,11 @@ defmodule Honeybadger.Utils do
     |> String.downcase()
   end
 
+  def uuid() do
+    :crypto.strong_rand_bytes(16)
+    |> Base.encode16(case: :lower)
+  end
```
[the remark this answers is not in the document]
rabidpraxis: Yeah, the name was wrong. 

The backend does not care about the format of the request id, but this matches the format Plug returns for it's generated request ids.

lib/honeybadger/insights/plug.ex
```diff
@@ -32,14 +32,35 @@ defmodule Honeybadger.Insights.Plug do
...
 
+  def get_telemetry_events() do
+    events = get_insights_config(:telemetry_events, @telemetry_events)
+
+    Enum.reduce(events, events, fn init, acc ->
+      if List.last(init) == :stop do
+        acc ++ [Enum.drop(init, -1) ++ [:start]]
+      else
+        acc
+      end
```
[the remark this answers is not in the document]
rabidpraxis: For all the events that end in :stop, it creates the same event but with :start at the end instead.

rabidpraxis:
I made some tweaks.

I took a step back and loaded some production like open source apps, since I needed a better understanding of real-world live view usage. Because of this, I have changed my strategy and am no longer generating a request_id per event, as it seems like many apps are much more chatty with events than I thought. Now we attempt to load a request_id from the process tree, or we create a new one if it does not exist.

Along with that I added a module to house the request id stuff (`Honeybadger.RequestId`). In it there is some code to hydrate the current process with any parent process's request_id (`Honeybadger.RequestId.inherit_or_initialize/1`). I think that passing the request_id in the state for all the things is the best method, but this is an alternative option that could require less work.

lib/honeybadger.ex
```diff
@@ -435,6 +437,25 @@ defmodule Honeybadger do
     :ok
   end
 
+  def set_request_id(request_id) do
+    Process.put(:hb_request_id, request_id)
+  end
+
+  def get_request_id do
+    Process.get(:hb_request_id)
+  end
```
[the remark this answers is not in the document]
rabidpraxis: Not really. I mean the desired result is that they would install the honeybadger client and it would track all the things automatically without any extra work, but I have found that is not the elixir way at all (unlike ruby where its easy, but full of footgun opportunities)

lib/honeybadger/insights/ecto.ex
```diff
@@ -62,6 +70,37 @@ defmodule Honeybadger.Insights.Ecto do
...
+      line = location[:line]
+
+      if file && line do
+        ["#{file}:#{line}", "#{inspect(module)}.#{function}/#{arity}"]
+      else
+        [nil, "#{inspect(module)}.#{function}/#{arity}"]
+      end
+    else
+      [nil, "#{inspect(module)}.#{function}/#{arity}"]
+    end
```
sorentwo, line 103: Can remove some duplication and flatten out the conditional:

```suggestion
    position =
      if is_list(location) and Keyword.has_key?(location, :file) do
        "#{location[:file]}:#{location[:line]}"
      else
        nil
      end

    [position, Exception.format_mfa(module, function, arity)]
```

lib/honeybadger/insights/oban.ex
```diff
@@ -44,10 +44,13 @@ defmodule Honeybadger.Insights.Oban do
...
       |> Oban.insert()
   """
-  def add_request_id(%Ecto.Changeset{} = changeset) do
-    meta = Ecto.Changeset.get_field(changeset, :meta) || %{}
-    updated_meta = Map.put(meta, "hb_request_id", Honeybadger.get_request_id())
-    Ecto.Changeset.change(changeset, meta: updated_meta)
+  def add_request_id(changeset) do
+    meta =
+      changeset
+      |> Ecto.Changeset.get_field(:meta, %{})
```
sorentwo, line 50: This _might_ require an `xref` exclude declaration in `mix.exs`:

```elixir
xref: [exclude: [Ecto.Changeset]]
```

lib/honeybadger/request_id.ex
```diff
@@ -0,0 +1,42 @@
+defmodule Honeybadger.RequestId do
+  @key __MODULE__
```
sorentwo, line 2: This is private and shouldn't show up in documentation:

```suggestion
defmodule Honeybadger.RequestId do
  @moduledoc false
  
  @key __MODULE__
```

sorentwo (approved):

lib/honeybadger/insights/ecto.ex
rabidpraxis: Oooh, handy!

PR #608, opened 2025-04-22
Add Events sample rate

rabidpraxis:
This adds a `Honeybadger.EventsSampler` module to supply sampling logic, keep track of sampled counts for logging, and is a good start for adaptive sampling ;)

### DIFF
--- a/README.md
+++ b/README.md
@@ -343,6 +343,25 @@ When explicit passing is needed:
 - Supervised processes
 - Processes started through other OTP abstractions
 
+### Event Sampling
+
+You can enable event sampling to reduce the number of events sent to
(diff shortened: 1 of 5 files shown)


stympy:
Our [Ruby example for sampling](https://docs.honeybadger.io/lib/ruby/getting-started/sending-events-to-insights/#sampling-events) uses an integer rather than a float... do you think one way is better than the other?

stympy (approved):

rabidpraxis:
> do you think one way is better than the other?

No. But I think we should be consistent, so what ever we choose will be the way all clients do it. Thinking about it more, perhaps whole percentage numbers are more friendly, so I'm in favor of using that.

PR #635, opened 2025-07-01
Move regex constants to functions

tjchambers:
OTP 28 does not allow regex in constants. Cnnverting them to simple functions.

cannot inject attribute @excluded_queries into function/macro because cannot escape #Reference<0.441870236.419561494.195505>. The supported values are: lists, tuples, maps, atoms, numbers, bitstrings, PIDs and remote functions in the format &Mod.fun/arity

### DIFF
--- a/lib/honeybadger/insights/ecto.ex
+++ b/lib/honeybadger/insights/ecto.ex
@@ -58,13 +58,6 @@ defmodule Honeybadger.Insights.Ecto do
   @required_dependencies [Ecto.Repo]
   @telemetry_events []
 
-  @excluded_queries [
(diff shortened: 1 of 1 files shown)


joshuap:
Thanks @tjchambers! We'll review next week when @rabidpraxis returns from vacation.

joshuap:
Oops, I missed the part where this was blocking your deploys @tjchambers. On second thought, I will review it right now. 😁

joshuap (approved):
Looks good, thanks!

joshuap:
Released in 0.24.1

PR #652, opened 2025-10-27
feat: ash Insights integration

rabidpraxis:


lib/honeybadger/insights/ash.ex
```diff
@@ -0,0 +1,186 @@
...
+      config :honeybadger,
+        ash_domains: [MyApp.Accounts, MyApp.Posts]
+
+  ## Custom Configuration
+
+  You can customize this module's behavior with the following configuration options:
+
+      config :honeybadger, insights_config: %{
+        ash: %{
+          # Additional custom telemetry events to listen for alongside auto-discovered ones
```
joshprice: Which ones would be auto-discovered?

